PR #5, opened 2019-07-07
Update package.json with types = dist/index.d.ts declaration

RAnders00:
Allows the typescript compiler to check types using the definitions from the module itself

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.2",
   "description": "Stop throwing errors, and instead return Results!",
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
     "test": "jest",
     "clean": "rm -rf ./dist",


gDelgado14:
Hey @RAnders00 sorry for the late response. I am not sure if this is necessary. I'm currently using runtypes inside of [this other project of mine](https://github.com/parlez-vous/server) and the ts compiler is able to infer types just fine without this option. 

Can you show me a reproducible setup where the ts compiler isn't picking up the types?

RAnders00:
In WebStorm at least the IDE does not provide any autocompletion and also tries to import from `neverthrow/dist`. I was actually not aware that tsc would be able to read `index.d.ts` nevertheless.

gDelgado14:
Ok got it. I'm inclined to add this! Just a last question, are there docs out there that discuss the `types` option in the package.json further?

RAnders00:
Definitely! This is a link to the typescript documentation directly: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

gDelgado14:
@RAnders00 version `1.0.3` is live on npm :)

PR #12, opened 2020-02-19
📝 README update match method

hugomartinet:
There was some deprecated doc for the match method (mentioned in this issue : https://github.com/gDelgado14/neverthrow/issues/11).

Also it was not specified that the return type is not a Result, meaning the match method is more a combination of _unsafeUnwrap and _unsafeUnwrapErr, rather than map and mapErr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,58 @@ const notNested = nested.andThen(innerResult => innerResult)
 
 ---
 
+### `Result.match` (method)
+
+Given 2 functions (one for the `Ok` variant and one for the `Err` variant) execute the function that matches the `Result` variant.
+
+Match callbacks do not necessitate to return a `Result`, however you can return a `Result` if you want to.
+
+**Signature:**
+```typescript
+match<A>(
+  okFn: (t:  T) =>  A,
+  errFn: (e:  E) =>  A
+): A => { ... }
+```
+
+
+`match` is like chaining `map` and `mapErr`, with the distinction that with `match` both functions must have the same return type.
+
+
+**Example:**
+
+```typescript
+const result = computationThatMightFail()
+
+const successCallback = (someNumber: number) => {
+  console.log('> number is: ', someNumber)
+}
+
+const failureCallback = (someFailureValue: string) => {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -349,6 +318,39 @@ const result = err(12)
...
+
+---
+
+### `Result.match` (method)
+
+Given 2 functions (one for the `Ok` variant and one for the `Err` variant) execute the function that matches the `Result` variant.
+
+Both functions will unwrap the `Result` object and must have the same return type.
+
+`match` is sort of like combining `_unsafeUnwrap` and `_unsafeUnwrapErr`.
```
gDelgado14: Thanks for the PR! 

> `match` is sort of like combining `_unsafeUnwrap` and `_unsafeUnwrapErr`.

Regarding this comment, note that the analogy is more accurate when you use `map` and `mapErr`. 

The reason is that `match` is a alternative api to chaining `map` and `mapErr`. 

```typescript
import { ok, Result } from 'neverthrow'

const myval: Result<number, string> = ok(100)

const successCallback = (someNumber: number) => {
  console.log('> number is: ', someNumber)
}

const failureCallback = (someFailureValue: string) => {
  console.log('> boooooo')
}

// method chaining api
// note that you DONT have to append mapErr
// after map which means that you are not required to do
// error handling
myval
  .map(successCallback)
  .mapErr(failureCallback)


// match api
// works exactly the same as above,
// except, now you HAVE to do error handling :)
myval.match(
  successCallback,
  failureCallback
)
```

You can play with the source here:

https://repl.it/repls/FrugalMajorRedundantcode

One tiny distinction between these two similar approaches is that with method chaining, your callbacks don't need to have the same return type, whereas with match you do!

So my ask is that you change that sentence to maybe include the above explanation, or something along those lines. I realize in retrospect that I should have been more clear when I wrote, "`match` is sort of like combining `map` and `mapErr`".

hugomartinet: Thanks for the feedback, and for clarifying the match method.
I'll apply this changes right away !

Issue #32, opened 2020-05-11
Treating async operations as first-class citizens


paduc:
Hi and thanks for writing this cool library !
I've been looking at monad libraries for a while now and they all seamed unnecessarily complex to me. I came to the conclusion that the main (only?) benefit of Either types was to have typed Errors. When I read your blog article, I thought "right on !".

There is one thing that bugs me though and it has to do with handling async operations.
Your (great) library offers a ```Result``` type with a bunch of great api methods like map, mapErr, andThen, match, etc. but as soon as an operation is async, there aren't as many options.
I like the ```chain``` solution but it doesn't offer the same chainability as we would get with the synchronous ```Result```.

I understand this has to do with the fact that you don't have control over the api of ```Promise<Result>``` but I believe there might be an elegant solution: having a wrapper around that Promise. A ```ResultAsync<T>``` type that would have more or less the same api as ```Result<T>``` and could offer full chainability. 

Furthermore, I was thinking that whatever the ```Result``` could do, the ```ResultAsync``` could do too. Therefor, I'm not even sure we need both types, only the async version. In the Real World, there's always an async operation somewhere in the chain anyway, right ?

The method signatures would look something like this:
```typescript


map<U>(fn: T => U | T => Promise<U>):  ResultAsync<U, E>
mapErr<U>(fn: E => F | E => Promise<F>):  ResultAsync <T, F>
andThen<U>(fn: T => ResultAsync <U,E>): ResultAsync <U, E>
match<A>(
  okFn: (t:  T) =>  A,
  errFn: (e:  E) =>  A
): Promise<A>

then(cb: (T | E) => any) // a method to make the ResultAsync a thenable which can have the result extracted by calling either .then() or await

```

And a code example:
```typescript

const newUserId = await api.checkAvailability()
                           .andThen(api.createNewUser)
                           .map({ id } => id);

// newUserId is of type string | Error
if(newUserId instanceof Error){
   // newUserId is an Error
   console.log(newUserId.message);
}
else{
   // newUserId is a string
   console.log("new user id is ", newUserId)
}
```

Thanks again and looking forward to your thoughts !

supermacro:
Hey @paduc,

Thanks for the kind words!

I like the idea, and this has been something that has crossed my mind in the past. So it's great to hear that others are thinking about it too.

I recently recalled that BlueBird implements its own Promise API that behaves much like native promises, except it adds a bunch more useful methods.

[...]

PR #34, opened 2020-05-13
Introduce a ResultAsync type for better async Results

paduc:
This is the PR for issue #32 

I'm waiting for your confirmation of the new API before updating the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,12 @@
 
 Encode failure into your program.
 
-This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).  
-It also offers a `ResultAsync` type which wraps a `Promise<Result>` to enable chaining asynchronous tasks.
+This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).
 
-`neverthrow` also exposes an api for chaining sequential asynchronous tasks ([docs below](#chaining-api)) in a more functionnal manner.
+For asynchronous tasks, `neverthrow` offers a `ResultAsync` class which wraps a `Promise<Result>` and enables chaining.  
+`ResultAsync` is `thenable` meaning it behaves exactly like a native `Promise<Result>`: the underlying `Result` can be accessed using the `await` or `.then()` operators.
+
+`neverthrow` also exposes `chain(...)` methods for chaining asynchronous tasks in a functional style ([docs below](#chaining-api)). However, these methods might be deprecated in the future. It is advised to use the `ResultAsync` instead.
 
 [Read the blog post](https://gdelgado.ca/type-safe-error-handling-in-typescript.html#title) which explains _why you'd want to use this package_.
 
@@ -70,13 +72,15 @@ type Result<T, E> = Ok<T, E> | Err<T, E>
 
 Asynchronous methods can return a `ResultAsync` type instead of a `Promise<Result>` in order to enable further chaining.
 
-This is useful for handling asynchronous apis like database queries, timers, http requests, ...
+`ResultAsync` is `thenable` meaning it behaves exactly like a native `Promise<Result>`: the underlying `Result` can be accessed using the `await` or `.then()` operators.
+
+This is useful for handling multiple asynchronous apis like database queries, timers, http requests, ...
 
 Example:
 
 ```typescript
 import { errAsync, ResultAsync } from 'neverthrow'
-import { insertIntoDb } from 'imaginary-database"
+import { insertIntoDb } from 'imaginary-database'
 // Let's assume insertIntoDb has the following signature:
 // insertIntoDb(user: User): Promise<User>
 
@@ -99,7 +103,9 @@ const asyncRes = addUserToDatabase({ name: 'Tom' }) // asyncRes is a `ResultAsyn
 const asyncRes2 = asyncRes.map((user: User) => user.name) // asyncRes2 is a `ResultAsync<string, Error>`
 
 // A ResultAsync acts exactly like a Promise<Result>
-// It can be transformed back into a Result just like a Promise would, using await or .then()
+// It can be transformed back into a Result just like a Promise would:
+
+// using await
 const res = await asyncRes
 // res is a Result<string, Error>
 if (res.isErr()) {
@@ -107,6 +113,16 @@ if (res.isErr()) {
 } else {
   console.log('Successfully inserted user ' + res.value)
 }
+
+// using then
+asyncRes.then(res => {
+  // res is Result<string, Error>
+  if (res.isErr()) {
+    console.log('Oops fail: ' + res.error.message)
+  } else {
+    console.log('Successfully inserted user ' + res.value)
+  }
+})
 ```
 
 ## Top-Level API
@@ -687,6 +703,9 @@ const resultMessage = await validateUser(user)
 # 🔗
 
 ## Chaining API
+
+> Disclaimer: the preferred solution to chaining asynchronous tasks is `ResultAsync`.  
+> The following method might be deprecated in the future.
 
 tldr: `chain` is the `.andThen` equivalent for `Result`s wrapped inside of a `Promise`.
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # NeverThrow 🙅
 
 [![Build Status](https://travis-ci.com/supermacro/neverthrow.svg?branch=master)](https://travis-ci.com/supermacro/neverthrow)
@@ -7,17 +6,16 @@
 
 Encode failure into your program.
 
-This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).
+This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).  
+It also offers a `ResultAsync` type which wraps a `Promise<Result>` to enable chaining asynchronous tasks.
 
-`neverthrow` also exposes an api for chaining sequential asynchronous tasks ([docs below](#chaining-api)).
+`neverthrow` also exposes an api for chaining sequential asynchronous tasks ([docs below](#chaining-api)) in a more functionnal manner.
 
-[Read the blog post](https://gdelgado.ca/type-safe-error-handling-in-typescript.html#title) which explains *why you'd want to use this package*.
-
+[Read the blog post](https://gdelgado.ca/type-safe-error-handling-in-typescript.html#title) which explains _why you'd want to use this package_.
 
 This package works for both JS and TypeScript. However, the types that this package provides will allow you to get compile-time guarantees around error handling if you are using TypeScript.
 
 `neverthrow` draws inspiration from [Rust](https://doc.rust-lang.org/std/result/enum.Result.html), and [Elm](https://package.elm-lang.org/packages/elm/core/latest/Result). It is also a great companion to [fp-ts](https://gcanti.github.io/fp-ts/).
-
 
 > Need to see real-life examples of how to leverage this package for error handling? See this repo: https://github.com/parlez-vous/server
 
@@ -27,11 +25,9 @@ This package works for both JS and TypeScript. However, the types that this pack
 > npm install neverthrow
 ```
 
-
-
-
-
 ## Usage
+
+### Synchronous API
 
 Create `Ok` or `Err` instances with the `ok` and `err` functions.
 
@@ -58,18 +54,60 @@ if (mappedYes.isOk()) {
 }
 ```
 
-
 `Result` is defined as follows:
 
 ```typescript
-type  Result<T, E>
-  =  Ok<T, E>
-  |  Err<T, E>
+type Result<T, E> = Ok<T, E> | Err<T, E>
 ```
 
 `Ok<T, E>`: contains the success value of type `T`
 
 `Err<T, E>`: contains the failure value of type `E`
+
+---
+
+### Asynchronous API
+
+Asynchronous methods can return a `ResultAsync` type instead of a `Promise<Result>` in order to enable further chaining.
+
+This is useful for handling asynchronous apis like database queries, timers, http requests, ...
+
+Example:
+
+```typescript
+import { errAsync, ResultAsync } from 'neverthrow'
+import { insertIntoDb } from 'imaginary-database"
+// Let's assume insertIntoDb has the following signature:
+// insertIntoDb(user: User): Promise<User>
+
+// We can create a synchronous method that returns a ResultAsync
+function addUserToDatabase(user: User): ResultAsync<User, Error> {
+  if (user.name.length < 3) {
+    // Throw a async result from a synchronous block thanks to the errAsync helper
+    return errAsync(new Error('Username is too short'))
+  }
+
+  // Wrap the async method into a ResultAsync thanks to fromPromise
+  // The seconds argument catches the error from the promise
+  return ResultAsync.fromPromise(insertIntoDb(user), () => new Error('Database error'))
+}
+
+// We can now call the method above
+const asyncRes = addUserToDatabase({ name: 'Tom' }) // asyncRes is a `ResultAsync<User, Error>`
+
+// We can chain the ResultAsync to build another ResultAsync (see full api below)
+const asyncRes2 = asyncRes.map((user: User) => user.name) // asyncRes2 is a `ResultAsync<string, Error>`
+
+// A ResultAsync acts exactly like a Promise<Result>
+// It can be transformed back into a Result just like a Promise would, using await or .then()
+const res = await asyncRes
+// res is a Result<string, Error>
+if (res.isErr()) {
+  console.log('Oops fail: ' + res.error.message)
+} else {
+  console.log('Successfully inserted user ' + res.value)
+}
+```
 
 ## Top-Level API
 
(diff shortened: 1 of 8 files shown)


src/result-async.ts
```diff
@@ -0,0 +1,131 @@
+import { Result, Ok, Err } from './result'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isPromise<T>(obj: any): obj is Promise<T> {
```
supermacro, line 4: Please use `unknown` instead of `any` for function argument. 

I think it would be beneficial to have a few tests for this function. Are there any edge cases that you think that should be tested?

I was initially thinking it would be beneficial to use a 3rd party library, but that would add unjustified bloat to this package. But now that there's a custom-made `isPromise` validator, it would be prudent to ensure it's working as expected (plus to verify regressions in the future).

paduc: Hey, this method was copied from this repo: [then/is-promise](https://github.com/then/is-promise).

The reason I didn't import it is because of the risks associated with dependencies. Very recently, the `is-promise` repo (which is listed as a dependency to 3.4M projects!) introduced a breaking change which wreaked havoc across a huge number of dependants. 

I think it's trivial enough to just paste here. This implementation coincides with the specification of a `thenable`: an `object` or `function` with a `then` property which is a `function`. It's a looser alternative to "instanceof Promise" which returns false for promises coming from other libs like Bluebird for instance. I don't see any edge-cases and if 3.4M projets use it, I guess there aren't any.

I'm not sure if having tests for this method is necessary to be honest. If you prefer, we can copy those located [here](https://github.com/then/is-promise/blob/master/test.js).

paduc: As for the `unknown` argument type, I tried but I have a compiler error on the `obj.then` section: `Property 'then' does not exist on type 'object'`.

paduc:
> Hey @paduc unfortunately I ran out of time this morning when reading through your PR, so I haven't really looked at much. Basically the only thing I've looked at thoroughly is `isPromise`. Please give me a few days in order to properly do code review.

No problem, take your time ! There aren't many lines of code but the implications aren't obvious... 

The principle is this: each `ResultAsync` instance holds a `Promise<Result>`.
For each chaining event (`map/mapErr/andThen`), the inner promise is chained with a `.then` which handles the different cases (Error/Value) and returns the appropriate `Result`, resulting in a new `Promise<Result>` that is then wrapped inside a new `ResultAsync` instance which is returned.

The `then` class method makes `ResultAsync` awaitable. It transforms a `ResultAsync` back into a `Promise<Result>`. 

One additional aspect: each chaining method accepts asynchronous function arguments. So the map function might return either a value (which is then wrapped in `Ok`) or a promise (which has to be unpacked with a `.then`).
A beauty of promises is that they are flattened automatically: returning an instance of `Result` or `Promise<Result>` inside a `then` callback will always result in a `Promise<Result>`.

Don't hesitate if you have further questions/remarks !

src/result-async.ts
supermacro: Yup, I agree it makes sense not to use a third-party lib. But for the sake of tracking regressions, there should definitely be some tests for this function.

I tried getting rid of `any`, and got as far as [this](https://www.typescriptlang.org/play/#code/C4TwDgpgBACgNgQwJYDsDyAjAVhAxsKAXigCU8B7AJwBMAeAZ2EtQHMAaKAVxQGsVyA7igB8AKFEAzbviTkUUJPXjIUAKXqYc+ABQA3BHABcXXvyEBKY-rgL6sRKk14CAb1FQolCME6V5oSHIJKGsiQmIAcnJsZwioADJ4kIMoAEJwqBROOBtRAF8ocUlpYFl5RRhKcgBbRQhaABVhbWisY24+QRRLKFbbWCra+nqmqDcPLx8-KG13DzTUvsS5j20KhzUNGJ1W8ygAH32oAIgg3uwwyKkUGTkIveX54-BT4NaAOmAACwh5cKuSmUInNzPlREA) ... but I am still getting type errors and so I think it's justified to keep the `any` in this case.

paduc: I agree about the tests. I just pushed a commit with extensive tests.

src/is-promise.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isPromise = <T>(obj: any): obj is Promise<T> => {
```
supermacro, line 2: For the sake of consistency with the rest of the package, please`export` here instead of exporting an object at the end of the file.

src/result-async.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class ResultAsync<T, E> {
+  private _promise: Promise<Result<T, E>>
+
+  constructor(res: Promise<Result<T, E>>) {
+    this._promise = res
+  }
+
+  static fromPromise<T, E>(promise: Promise<T>): ResultAsync<T, E> {
+    return new ResultAsync(promise.then((value: T) => new Ok<T, E>(value)))
```
supermacro, line 12: What happens if calling `promise.then` results in a rejected promise? It would be odd if neverthrow started throwing unhandled exceptions 🙈 !
  - What if there was a `catch` here that returned an `Err`?

src/result-async.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+
+      const newValue = _f(res.value)
+
+      if (isPromise(newValue)) {
+        return newValue.then(value => new Ok<A, E>(value))
+      }
+
+      return new Ok<A, E>(newValue)
+    })
```
supermacro, line 28: Looks great! I have a few thoughts. 

For the sake of keeping code as legible as possible:
- Any reason not to use `async` / `await` syntax? It might lead to less indentation
- Please omit explicit type annotations if the compiler can infer the type already

The code on line 16 would look like this:  

```typescript
    const newPromise = this._promise.then((result) => { ...
```

Secondly, the instantiation of `Err` and `Ok` values on line 18, 24 and 27 shouldn't need to be annotated with their respective type arguments either. See if you can omit them. The compiler should be able to infer the types in all of these cases. 

If you do go ahead with these changes, please apply them to all the methods below as well! Otherwise please respond with some counterarguments so we can agree on a consistent coding style!

src/result-async.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    return new ResultAsync(newPromise)
+  }
+
+  match = <A>(ok: (t: T) => A, _err: (e: E) => A): Promise<A> => {
+    return this._promise.then(res => res.match(ok, _err))
+  }
+
+  // Makes ResultAsync awaitable
+  then(_successCallback: (res: Result<T, E>) => void) {
```
supermacro, line 74: Cool! I learned something new :) By the way, does this method need to return anything? Seems counter-intuitive that this method doesn't return a value. 

In my head I am thinking, "in normal promises, I can chain `.then` as many times as I want ... so wouldn't that imply that a `.then` method returns a `Promise`?"

src/result-async.ts
```diff
@@ -0,0 +1,123 @@
...
+    this._promise.then(_successCallback)
+  }
+}
+
+export const okAsync = <T, E>(value: T): ResultAsync<T, E> =>
+  new ResultAsync(Promise.resolve(new Ok<T, E>(value)))
+
+export const errAsync = <T, E>(err: E): ResultAsync<T, E> =>
+  new ResultAsync(Promise.resolve(new Err<T, E>(err)))
```
supermacro, line 83: What's the use case for having `okAsync` and `errAsync`? It seems to me that if I want to convert a normal result into a async result, I would call `.asyncMap`, no?

I would prefer to keep the surface area of code as small as possible. This is where I'm coming from when I ask this question, and If there's no good reason right now to keep `okAsync` and `errAsync`, then I would ask that it be removed.

src/result-async.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  constructor(res: Promise<Result<T, E>>) {
+    this._promise = res
+  }
+
+  static fromPromise<T, E>(promise: Promise<T>): ResultAsync<T, E> {
+    return new ResultAsync(promise.then((value: T) => new Ok<T, E>(value)))
+  }
+
+  map<A>(_f: (t: T) => A | Promise<A>): ResultAsync<A, E> {
```
supermacro, line 15: change `_f` to `f` since it is being used in the body of this method. The convention that I've been abiding by is that if an argument is not being used then I signal it to the reader (and the compiler) by prepending the underscore.

Please follow this convention for all code in this PR.

Example:

https://github.com/supermacro/neverthrow/blob/master/src/result.ts#L77

supermacro:
Hey @paduc ... not sure why the TravisCI status check didn't run, but I have now resolved that issue. 

You can ignore the Codacy status check (it's something I'm still evaluating and trying out). But the TravisCI check is required to pass before merging!

src/result-async.ts
paduc: You are totally right and I actually had already modified this function in another branch.
What I propose is to offer a second argument to `fromPromise` to handle the error case. This second argument should be a function which receives whatever is thrown by the promise and returns an `E` type value.

It looks like this now:
```typescript
  static fromPromise<T, E>(promise: Promise<T>, errorFn?: (e: unknown) => E): ResultAsync<T, E> {
    let newPromise: Promise<Result<T, E>> = promise.then((value: T) => new Ok(value))
    if (errorFn) {
      newPromise = newPromise.catch(e => new Err<T, E>(errorFn(e)))
    }
    return new ResultAsync(newPromise)
  }
```

src/result-async.ts
paduc: > * Any reason not to use `async` / `await` syntax? It might lead to less indentation

These methods need to be remain synchronous (as they are returning a `ResultAsync` instance) and `async` / `await` is only available inside asynchronous functions.

However, it's possible to pass an asynchronous function to `.then()` and `await` the value instead of chaining with an addition `.then()`. 
I also noticed, we can `await` non-promise values as well. Given we don't know if the provided function returns a promise or not, we can `await` in both cases.

Looks a lot cleaner like this:
```typescript
  map<A>(f: (t: T) => A | Promise<A>): ResultAsync<A, E> {
    const newPromise = this._promise.then(async (res: Result<T, E>) => {
      if (res.isErr()) {
        return new Err<A, E>(res.error)
      }

      return new Ok<A, E>(await f(res.value))
    })

    return new ResultAsync(newPromise)
  }
```

Added bonus: we don't need `isPromise` at all anymore 🎉

src/result-async.ts
paduc: Of course, I misinterpreted your conventions. I fixed this.

src/result-async.ts
paduc: Here's an example use-case:

```typescript
const createUser = (login: string): ResultAsync<boolean, Error> => {
  if (login.length < 3) {
    return errAsync(new Error('Username is too short'))
  }

  return ResultAsync.fromPromise(
    insertUserIntoDb(login),
    () => new Error('Failed to insert into database'),
  )
}
```

=> A method that should return a `ResultAsync` and also performs synchronous operations. It's the async equivalent of `Result`s `ok` and `err` handlers.
It avoids writing `return new ResultAsync(Promise.resolve(new Err<T, E>(err)))`...

src/result-async.ts
paduc: You are absolutely correct.

I added a test (which of course didn't pass with the previous version) and fixed the implementation.

src/result-async.ts
```diff
@@ -0,0 +1,70 @@
...
+    this._promise = res
+  }
+
+  static fromPromise<T, E>(promise: Promise<T>, errorFn?: (e: unknown) => E): ResultAsync<T, E> {
+    let newPromise: Promise<Result<T, E>> = promise.then((value: T) => new Ok(value))
+    if (errorFn) {
+      newPromise = newPromise.catch(e => new Err<T, E>(errorFn(e)))
+    }
+    return new ResultAsync(newPromise)
+  }
```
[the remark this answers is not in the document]
paduc: Good idea !

It's a bit tricky though because `process.env` is only for `node` runtimes. Isn't this lib also for the browser ?

We could either add a condition to check if it's a browser environment (would we want to log in the browser as well ?).

We could also make the second argument mandatory.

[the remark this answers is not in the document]
supermacro: @paduc good point. I was thinking that `process.env` would be available because of webpack or parcel.

I think this warning should be logged anywhere. The reason I didn't want to log in production is because the user might already be dealing with promise rejections on their own. So it would be annoying to keep logging the warning even though it's irrelevant to them.

I don't think making it mandatory makes sense because users may prefer to catch the exception and return an `Err` themselves.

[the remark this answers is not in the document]
paduc: Ok to recap, the solution would be:
- In node, only log outside of prod and test environments
- In the browser, always log ?

What if we only log if the second argument is `undefined` (meaning, no warning if the user provides `null` for instance ?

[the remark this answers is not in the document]
supermacro: Yes, let's do that. Log outside of tests or prod for node. Always log in a browser environment (although some front end apps might have `process.env` defined, like webpack projects).

Don't add the `undefined` vs `null` functionality. I don't think it's worth it.

paduc:
Hello @supermacro,

I'm working on the documentation and I am wondering about the `chain` methodology.
Obviously, you'll want to keep it to avoid breaking changes and also because it's a nice alternative to chaining.

However, you might want to address the fact that there are two ways of handling asynchronous behaviour now.

README.md
```diff
@@ -385,15 +425,267 @@ import { parseHeaders } 'imaginary-http-parser'
...
+myResult.isErr() // true
+```
 
 ---
+
+### `ResultAsync.fromPromise` (method)
+
+Transforms a `Promise<T>` into a `ResultAsync<T, E>`.
+
+The second argument handles the rejection case of the promise. If it is ommited, the code might throw.
```
supermacro, line 491:
```suggestion
The second argument handles the rejection case of the promise. If it is ommited, **the code might throw** because `neverthrow` does not know if the promise you are passing to `fromPromise` has any promise rejection logic associated to it (via a `.catch` method call or `catch (err) {}` block).
```

supermacro:
Hey @paduc almost there!

I think what I'll do is remove the `chain` api & documentation in a breaking-change release when I get some time.

I don't know when (or if) I will do that, but that's what I am thinking. So right now it's best to leave the `chain` docs as they are.

I noticed that travis / github are bugging out, so I'll simply run the CI checks on my machine after you've pushed those last documentation changes I requested, and if they pass I can merge and then publish a new version to npm 😁

Thanks for your patience! If you have any feedback on the whole process, feel free to share.

paduc:
> Hey @paduc almost there!

Awesome ! Can't wait !

> 
> I think what I'll do is remove the `chain` api & documentation in a breaking-change release when I get some time.
> 
> I don't know when (or if) I will do that, but that's what I am thinking. So right now it's best to leave the `chain` docs as they are.
> 

Maybe we can add a line advising the use of `ResultAsync` and that it might be deprecated in the future ?

> I noticed that travis / github are bugging out, so I'll simply run the CI checks on my machine after you've pushed those last documentation changes I requested, and if they pass I can merge and then publish a new version to npm 😁
> 
> Thanks for your patience! If you have any feedback on the whole process, feel free to share.

This is my first meaningful open source contribution so I'd say the process is exciting and taught me a lot. I feel you've done a great job communicating your requirements and providing pointers !

README.md
```diff
@@ -58,19 +54,61 @@ if (mappedYes.isOk()) {
...
 
 `Ok<T, E>`: contains the success value of type `T`
 
 `Err<T, E>`: contains the failure value of type `E`
 
+---
+
+### Asynchronous API
+
+Asynchronous methods can return a `ResultAsync` type instead of a `Promise<Result>` in order to enable further chaining.
```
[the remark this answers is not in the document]
paduc: I added a line to this effect. I also added an example using `.then()` below.

README.md
```diff
@@ -7,18 +6,17 @@
 
 Encode failure into your program.
 
-This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).
-
-`neverthrow` also exposes an api for chaining sequential asynchronous tasks ([docs below](#chaining-api)).
+This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).  
+It also offers a `ResultAsync` type which wraps a `Promise<Result>` to enable chaining asynchronous tasks.
```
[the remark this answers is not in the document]
paduc: I added a line to this effect. What do you think ?

supermacro:
Hey @paduc,

There's good news and bad news haha .. I approved the PR, but when I went to run the CI checks (which are `npm run lint`, `npm run typecheck` and `npm run test`), they unfortunately failed because of a missing `node` types package. See screenshot below. 

Screenshot from 2020-05-20 08-28-55

Please ensure that the above three npm scripts pass successfully on your machine.

Hopefully by running `npm install --save-dev @types/node` all your checks will pass and we can merge today.

paduc:
> Hey @paduc,
> 
> There's good news and bad news haha .. I approved the PR, but when I went to run the CI checks (which are `npm run lint`, `npm run typecheck` and `npm run test`), they unfortunately failed because of a missing `node` types package. See screenshot below.
> 

My bad ! I had installed the `@types/node` locally but didn't commit the changes to `package.json` because I wasn't sure if we would keep the change yet.

I pushed a fix and all three commands run on my side.

supermacro:
👍

paduc:
🍾🥂

PR #41, opened 2020-05-22
Fix strictFunctionTypes error

### LINKED ISSUE #39: "strictFunctionTypes" compiler option leads to error
This is a continuation from this comment from https://github.com/supermacro/neverthrow/issues/36#issuecomment-631706079

Looks like the `strictFunctionTypes` compiler option leads to a rather cumbersome type error which looks to have merit. 

This [SO](https://stackoverflow.com/questions/51767338/what-is-the-benefit-of-using-strictfunctiontypes-in-typescript) thread and the [official docs on the subject](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html) are great resources for explaining what `strictFunctionTypes` is, but I personally have not sat down to really think this through to know what the implications for `neverthrow` are. 

Below is a demonstration where there are type errors when the `strictFunctionTypes` is turned on.

Code example: https://repl.it/repls/SparseSaltyPdf

Screenshot:

Screen Shot 2020-05-21 at 10 16 01 AM

paduc:
Relates to #39 : will remove the compiler error when using `strictFunctionTypes` but might not solve the fundamental type checking issue.

### DIFF
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -30,7 +30,7 @@ export class ResultAsync<T, E> {
     return new ResultAsync(newPromise)
   }
 
-  map<A>(f: (t: T) => A | Promise<A>) {
+  map<A>(f: (t: T) => A | Promise<A>): ResultAsync<A, E> {
     return new ResultAsync(
       this._promise.then(async (res: Result<T, E>) => {
         if (res.isErr()) {
@@ -42,7 +42,7 @@ export class ResultAsync<T, E> {
     )
   }
 
-  mapErr<U>(f: (e: E) => U | Promise<U>) {
+  mapErr<U>(f: (e: E) => U | Promise<U>): ResultAsync<T, U> {
     return new ResultAsync(
       this._promise.then(async (res: Result<T, E>) => {
         if (res.isOk()) {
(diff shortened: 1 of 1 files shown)


src/result-async.ts
```diff
@@ -30,8 +30,8 @@ export class ResultAsync<T, E> {
     return new ResultAsync(newPromise)
   }
 
-  map = <A>(f: (t: T) => A | Promise<A>) =>
-    new ResultAsync(
+  map<A>(f: (t: T) => A | Promise<A>) {
```
supermacro, line 33: While we're at it, would you be able to also specify the return type for all of these methods?

paduc: I believe I removed them at your request because the compiler could extrapolate them. Maybe I misunderstood.

Is it necessary if the compiler is capable of extrapolating them ?

supermacro:
I'm starting to believe that there's a problem with my configuration of travis ci 🤔 ... for now please run the checks locally.

paduc:
Should we change the `tsconfig.json` to add `strict` to catch this type of error at the `npm run typecheck` phase ?

src/result-async.ts
supermacro: Hmm I don't believe I requested to remove the return type. If I did I apologise.

like you mentioned, It's not necessary as the compiler can infer the return type. But by having an explicit type annotation, you create a contract. And in the future if another person comes in and changes this code, then it must adhere to the contract! And we can tell if the change still adheres to the contract or not via typechecking.

supermacro:
> Should we change the tsconfig.json to add strict to catch this type of error at the npm run typecheck phase ?

Not until it's deemed that the flag should always be set in a TS project, based on the ongoing discussion on #39

src/result-async.ts
paduc: Ok thanks for clearing this up. I just pushed a commit to add the return types back.

Issue #38, opened 2020-05-20
Simplify / organize docs


supermacro:
I am seeking suggestions on how to make the documentation easier to discover and learn.

I am considering using gitbook.com ... Any other suggestions?

paduc:
Gitbook is great. I'll be happy to lend a hand.

I think one improvement would be to start right away a real use-case example, to show the pain-point and the solution `neverthrow` offers.

Also, now that the api is pretty long, I would add a simple list of all methods signatures and a link to their specific paragraph. I think by looking at the signatures, you gets a sense of which method you're going to need.

supermacro:
> I think one improvement would be to start right away a real use-case example, to show the pain-point and the solution neverthrow offers.

I think that linking to my blog post is the right thing to do for those that aren't familiar with the purpose of this library. I wouldn't want to add a whole section explaining "why" within the README.

> Also, now that the api is pretty long, I would add a simple list of all methods signatures and a link to their specific paragraph. I think by looking at the signatures, you gets a sense of which method you're going to need.

Yeah that's a great idea. 

Something like this maybe?

----

### API Overview

- `Result` (Synchronous API [#](#link-to-syncronous-api))
  - `ok` 
  - `err`
  - `.map`
  - `.mapErr`
  - ...
  - ...
- `ResultAsync` (Asynchronous API [#](#link-to-asyncronous-api))
  - `.fromPromise` (static)
  - `okAsync`
  - `errAsync`
  - `.map`
  - ...
  - ...


### `Result` - Synchronous API

blah blah blah 

....


### `AsyncResult` - Asynchronous API

blah blah blah 

...

paduc:
Yes, exactly, a sort of "Table of contents". 
I would add two things:
- A pseudo type signature (not necessarily typescript but enough to get "Oh map transforms the value type" really fast)
- A link for each method

Like this maybe:

## API Overview

 * `Result` (Synchronous API [#](#link-to-syncronous-api))
   
   * [`ok(T) => Ok<T,?>`](#link-to-specific-api)
   * [`err(E) => Err<?,E>`](#link-to-specific-api)
   * [`.map(T => U) => Result<U,E>`](#link-to-specific-api)
   * [`.mapErr(E1 => E2) => Result<T,E2>`](#link-to-specific-api)
   * ...
   * ...
 * `ResultAsync` (Asynchronous API [#](#link-to-asyncronous-api))
   
[...]

PR #44, opened 2020-05-29
Termination operations

xxlabaza:
Hi!

I've added the so-called `termination` methods for returning a real value (or throwing an `Error` in case of `Err`).

The methods are useful when you want to **fail fast**, like here:

```typescript
// the method `get(string)` returns `Result<any, Error>`
const value = from(json).get('user.name').orError()
``` 

In the example above I don't want to handle an invalid value, I just want to get a proper result or throw an exception (in this case, it specified as a `Err` part).

Another example is for default values, where I want to use a default value if `Result` has `undefined` or `Err` value, like here:

```typescript
const value = from(json).get('user.name').or('Artem')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,8 +501,8 @@ err('error message').orError((errorResult) => new Error(errorResult))
 // throw new Error('Ooops!')
 err(new Error('Oooops!')).orError()
 
-// INVALID USAGE, unsupported operation
-// throw new Error("Invalid 'orError' usage. Error function was not specified and error type is not Error.")
+// throw new Error('error message'),
+// because the error object casted to string
 err('error message').orError()
 ```
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Encode failure into your program.
 
 This package contains a `Result` type that represents either success (`Ok`) or failure (`Err`).
 
-For asynchronous tasks, `neverthrow` offers a `ResultAsync` class which wraps a `Promise<Result>` and enables chaining.  
+For asynchronous tasks, `neverthrow` offers a `ResultAsync` class which wraps a `Promise<Result>` and enables chaining.
 `ResultAsync` is `thenable` meaning it behaves exactly like a native `Promise<Result>`: the underlying `Result` can be accessed using the `await` or `.then()` operators.
 
 `neverthrow` also exposes `chain(...)` methods for chaining asynchronous tasks in a functional style ([docs below](#chaining-api)). However, these methods might be deprecated in the future. It is advised to use the `ResultAsync` instead.
@@ -45,7 +45,7 @@ const yesss = ok(someAesomeValue)
 const mappedYes = yesss.map(doingSuperUsefulStuff)
 
 // neverthrow uses type-guards to differentiate between Ok and Err instances
-// Mode info: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types
+// More info: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types
 if (mappedYes.isOk()) {
   // using type guards, we can access an Ok instance's `value` field
   doStuffWith(mappedYes.value)
@@ -65,6 +65,24 @@ type Result<T, E> = Ok<T, E> | Err<T, E>
 `Ok<T, E>`: contains the success value of type `T`
 
 `Err<T, E>`: contains the failure value of type `E`
+
+It's also possible to terminate a `Result` with a default value or throw an `Error` in-place:
+
+```javascript
+// returns 42, if result is err or undefined
+result.or(42)
+
+// invokes calculateMyDefaultValue(), if result is err or undefined
+result.orGet(() => calculateMyDefaultValue())
+
+// throws the Error, if result is err
+result.orError(() => new Error('Ooops!'))
+
+// throws the Result's error, if result is err
+const result: Result<any, Error>
+...
+result.orError()
+```
 
 ---
 
@@ -319,7 +337,7 @@ parseResult.isErr() // true
 
 Same idea as `map` above. Except you must return a new `Result` or `ResultAsync`.
 
-If the provided method returns a `Result` the returned value will be a `Result`.  
+If the provided method returns a `Result` the returned value will be a `Result`.
 If the provided method returns a `ResultAsync` the returned value will be a `ResultAsync`.
 
 This is useful for when you need to do a subsequent computation using the inner `T` value, but that computation might fail.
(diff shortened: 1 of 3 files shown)


xxlabaza:
...aaand yes, I added ‘throw’ in **neverthrow** library)))

supermacro:
Hey @xxlabaza,

Thanks for opening a PR!

It looks like you are looking for `_unsafeUnwrap` and `_unsafeUnwrapErr` which do exactly what you're looking for. 

```typescript
ok(123)._unsafeUnwrap() // -> 123
ok(123)._unsafeUnwrapErr() // -> throws exception
err('abc')._unsafeUnwrap() // -> throws exception
err('abc')._unsafeUnwrapErr() // -> 'abc'
```

This api is not documented because a subsequent release of `neverthrow` improved the `isErr` and `isOk` api to provide type information to the TS compiler. Hence it was decided to not inform users of this available api. 

However, I see that there are valid use cases for early termination behaviour. Hence, I will document this API once again as part of https://github.com/supermacro/neverthrow/issues/38.

Please close this PR if you agree and are happy with the current situation!

xxlabaza:
Hi!

`_unsafeUnwrap*` methods are not the same as the suggested by me syntactically and logically. 

The **fail fast** example:

```typescript
// it always throws uninformative 'Called `_unsafeUnwrap` on an Err',
// instead of my own error object, specified in get(string) method. 
// Moreover, I loose the stack trace of the error here.
const value = from(json).get('user.name')._unsafeUnwrap() 
```

and I don't understand how to implement the `give me default value if error occurred` logic (my second example) in one line and without try-catch, with `_unsafeUnwrap*` methods.

Also, `or*` methods look more prettier)

paduc:
Hey @xxlabaza,

I like the idea of `or*` methods but I'm wondering if they're enabling bad practices (i.e. ignoring the error cases).

I guess the equivalent of `or` would be
```typescript
const res = from(json).get('user.name')
const value = res.isOk() ? res.value : 'Artem'
```

Why have a `get` method return a `Result` if you're not handling the error case ?

and `orError` 
```typescript
const res = from(json).get('user.name')
if(res.isErr()){ throw res.error }
// Use res.value as value
```

Where would you catch this error ? Or is it the point to have an unhandled exception ?

xxlabaza:
@paduc your examples take minimum 2 LOC (but normally should be 3-4 LOC, because of proper formatting ternary operator and if statement) and don't improve readability of the code. 

Moreover, imagine the case where you need to write several almost identical expressions:

```typescript
constructor (json: MyWrapper) {
  this.id = json.getInteger('id').orError();
  this.description = json.getString('description').or('');
  ...
}
``` 

Here, I don't want to handle the exception or do some extra logic. I want to write the reason, something like *"The JSON {} doesn't have field 'id'"*, as a message for Error, and throw it up through the call stack, where controller's try-catch statement put the reason in 400 response.

That is it. Extra readable and super useful!)

...and, as I understand, you suggest me to make it this way:

```typescript
constructor (json: MyWrapper) {
  const result1 = json.getInteger('id');
  if (result1.isErr()) {
    throw result1.error;
  } else {
    this.id = result1.value;
  }

  const result2 = json.getString('description');
  if (result2.isErr() || result2.value === undefined) {
     this.description = '';
  } else {
    this.description = result2.value;
  }
  ...
}
```

...doesn't look sexy)

paduc:
@xxlabaza I'm sorry I misread your answer to Gio and was trying to answer "how to implement this using the current api". My answer didn't take into account "in one line and without try-catch".

Thank you for clarifying your need. I believe there is a better way to handle it.

The way I see this, you have N calls to methods that return a `Result` and you would like handle them all together, like if they were one big `Result`. 
Using try/catch and throwing individually is one way of doing it but defeats the purpose of `neverthrow` which was made to handle errors without try/catch.

Here's a possible solution:
```typescript
// Start with a Record<string, Result<*, Error>
const recordOfResult: Record<string, Result<any, Error>> = {
  id: json.getInteger('id'),
  description: json.getString('description').match(
    d => ok(d), // In case of success, leave as is
    e => ok(''), // In case of error, transform into Ok('')
  ),
}

// Transform into a Result<Record<string, any>, Error>
const resultOfRecord = Object.entries(recordOfResult).reduce(
  // outerResult is the Result<Record<string, any>, Error> we are building
  // innerResult is Result<any, Error> that was returned by getInteger/getString
  (outerResult, [key, innerResult]) =>
    // andThen fn will be called only if outerResult is Ok
    outerResult.andThen(outerRecord =>
      // map fn will be called only if innerResult is Ok
      innerResult.map(innerValue => ({
        ...outerRecord,
        [key]: innerValue,
      })),
    ),
  ok({}),
)

// Controller can handle the success/error cases
if(resultOfRecord.isErr()){
  // Handle the error
  res.status(400).send({ error: resultOfRecord.error })
}
else{
  // Handle the success
  res.send(resultOfRecord.value)
}
```

Ok, it's not very readable. There might be a better way. Or maybe we could imagine `combine` or `zip` helpers to handle the cases where we want to combine multiple `Result` into a single one.

supermacro:
I think there are a few things being asked here.

Let me reiterate to make sure we are all on the same page:

@xxlabaza is requesting the following:

- Custom behaviour over `_unsafeUnwrap` and `_unsafeUnwrapErr`, but ultimately preserving the ability to fail (i.e. throw) if the `Result<T, E>` does not contain the `Ok` or `Err` variant that you are trying to "unwrap" or "get"
  - Ability to instantiate your own error so that you may have finer grained control over the stack trace
  - Ability to provide more informative information compared to the vague, "Called _unsafeUnwrap*" on an *
- Additional functionality to provide default values
  - When a `Result` is a `Ok` variant but you wanted the value inside of the `Err`
  - When a `Result` is a `Err` variant but you wanted the value inside of the `Ok`

I like these suggestions. And I am open to adding new methods to the `Result` and `ResultAsync` objects.

These are the methods that I am thinking are missing:

- `unwrapOrThrow<F extends Error>(customError: F): T`
  - Exact same behaviour of `_unsafeUnwrap` except that the user can provide a custom `Error` object that will be used to throw the exception when the `Result` is a `Err` variant
- `unwrapErrOrThrow<F extends Error>(customError: F): E`
  - Exact same behaviour of `_unsafeUnwrapErr` except that the user can provide a custom `Error` object that will be used to throw an exception when the `Result` is a `Ok` variant
- `unwrapOr(defaultVal: T) => T`
  - Unwraps a `Result<T, E>` and returns the inner `T` value of an `Ok`, or returns the provided default value if the `Result` is a `Err` variant 

Let me know what you think!

supermacro:
Hey @xxlabaza, any updates on this? I may close this PR soon. However, I do like the idea of adding these new api methods that I suggested.

xxlabaza:
oh, sorry, I forgot to close the PR. I already made my own implementation of the library)

PR #60, opened 2020-06-14
ESM Build Target Using Rollup

### LINKED ISSUE #58: Publish ESM Version
I think it would be beneficial to publish a ES module version alongside the commonjs version using the `module` field in the `package.json`. I've done this once with another package I made and I used `rollup` to build both versions. If you think this is a good idea, I could make a PR?

jsmith:
Closes #58 

## Details
- Uses [rollup](https://rollupjs.org/) along with [`rollup-plugin-typescript2`](https://github.com/ezolenko/rollup-plugin-typescript2) to build to two target files (`index.es.js` and `index.cjs.js`). This means that there is only a *single* file now for each target rather than 4 files (ie. `index.js`, `chain.js`, etc.).
- Rollup has their own official [`@rollup/plugin-typescript`](https://www.npmjs.com/package/@rollup/plugin-typescript) plugin but it does not have the best TS declarations support atm (see https://github.com/rollup/plugins/issues/394, https://github.com/rollup/plugins/issues/254, https://github.com/rollup/plugins/issues/243). Until these are resolved, I choose to just use `rollup-plugin-typescript2`.
- `rollup-plugin-typescript2` generates `index.d.ts`, `chain.d.ts`, `result.d.ts` and `result-async.d.ts` but this is inconsistent with the generated JavaScript files. This isn't a huge issues unless someone tries to import `neverthrow/dist/chain` which would error because the underlying `.js` doesn't exist. To remedy this issue, I used [`rollup-plugin-dts`](https://www.npmjs.com/package/rollup-plugin-dts) to merge `*.d.ts` files into a single `index.d.ts`.
- Because the build process is kinda two steps (first using `rollup-plugin-typescript2` and then using `rollup-plugin-dts`), I first write to `tmp/`, then write the merged `index.d.ts` to `dist/` and then moved `index.es.js` and `index.cjs.js` to `dist/`.
- I'm a bit sad that it's this complicated to generate two build outputs but once some of the issues I've linked to above are resolved this process will become much easier :)

## Publish Dry Run
```
➜ npm publish --dry-run
npm notice 
npm notice 📦  neverthrow@2.3.2
npm notice === Tarball Contents === 
npm notice 1.1kB  LICENSE          
npm notice 14.5kB dist/index.cjs.js
npm notice 14.2kB dist/index.es.js 
npm notice 1.5kB  package.json     
npm notice 25.3kB README.md        
npm notice 4.3kB  dist/index.d.ts  
npm notice === Tarball Details === 
npm notice name:          neverthrow                              
npm notice version:       2.3.2                                   
npm notice package size:  12.2 kB                                 
npm notice unpacked size: 60.8 kB                                 
npm notice shasum:        e90c96355a65c7e6bfa6664fa9e3a12e02693424
npm notice integrity:     sha512-8m34VsIllfdFN[...]E60K4SZJQLueg==
npm notice total files:   6                                       
npm notice 
+ neverthrow@2.3.2
```


supermacro:
Hey @jsmith thank you for opening this PR!

A few thoughts before merging. 

- Looks like typechecking is failing. Do you have any idea why that may be? 
- RE: Your detailed context & explanations on usage of various packages: Could you add this to the top of the `rollup.config.js` file as a module comment?

```typescript
/**
 * info info info info
  */
```

jsmith:
Ok just added the module comment. Also the type issues stemmed from changing `target` from `CommonJS` to `ES2015`. This [comment](https://github.com/babel/babel/issues/10237#issuecomment-513028440) explains it well but apparently changing the `target` also changed the module resolution algorithm. `rollup` requires a `target` >= `ES2015` btw which is why I made that change.

PR #70, opened 2020-06-30
feat(mapOr): Implement mapOr method

pawndev:


supermacro:
Hey @pawndev,

Thanks for opening a PR. 

I like the concept of unwrapping a `Result` and providing a default value. However, I do not agree with the API you are proposing. 

Please see [this](https://github.com/supermacro/neverthrow/pull/44#issuecomment-636351471) thread / comment that illustrates what my thoughts are on a more intuitive API. 

Specifically, in the thread I make reference to a `unwrapOr` method:

`Result<T, E>.unwrapOr(defaultVal: T) => T`

So your `mapOr` could be re-implemented as follows:

```typescript
import { err } from 'neverthrow

const myResult = err('Oh noooo')

const multiply = (val: number): number => number * 2

const unwrapped: number = myResult
  .map(multiply)
  .unwrapOr(10)
```

I would happily accept a PR for this method.

pawndev:
@supermacro like this: https://github.com/supermacro/neverthrow/pull/72 ?
We can switch to the other PR, and close this one if you want.

PR #72, opened 2020-06-30
feat(unwrapOr): Implement the unwrapOr method.

pawndev:


README.md
```diff
@@ -315,6 +315,30 @@ parseResult.isErr() // true
 
 ---
 
+### `Result.unwrapOr` (method)
+
+Unwrap the `Ok` value, or return de default if there is an `Err`
```
paduc, line 320: typo here on `or return _de_ default`

paduc:
This would definitely be a welcome addition for me !

README.md
pawndev: 🙈 I will change this. Thanks !

paduc: Simply unacceptable. 😉

supermacro:
Awesome! Very exciting addition to `neverthrow`. 

Thank you @pawndev for the contribution, and thanks @paduc for the second pair of eyes!

supermacro:
It's live!

https://github.com/supermacro/neverthrow/releases/tag/v2.6.0

I'm thinking it would make sense to also add this method to `ResultAsync`.

Do either of you have time to implement this?

```
ResultAsync<T, E>.unwrapOr(defaultVal: T) => Promise<T>
```

paduc:
@supermacro I thought about it but came to the conclusion that it is probably not necessary. When using a `ResultAsync` unwrapping always implies waiting for the promise.

```typescript
const resultAsync: ResultAsync<number, Error> = getMyNumber();

const myNumber  = (await resultAsync).unwrapOr(10);
```

Ok so we could avoid the parens with a `ResultAsync.unwrapOr` => `const myNumber = await resultAsync.unwrapOr(10);` but is it worth it ?

pawndev:
@supermacro awesome ! Thanks for merging/tagging it so fast !

@paduc I agree with you, but I'll take a look to see if it's worth to implement it

supermacro:
The reason I like the idea of having it for `ResultAsync` is because I see `ResultAsync` as a **near-exact** copy of `Result` ... but just asynchronous. 

So in my "vision" for `neverthrow`, I always want to have the same methods on `ResultAsync` as there are for `Result`. That way, there is zero (or close to zero) mental overhead when context-switching between `Result`s and `ResultAsync`s.

paduc:
Great vision indeed ! I just opened a new PR with this small addition.

PR #83, opened 2020-07-15
removes unneeded lint disabling

naseemkullah:
I believe these can now be removed as linter does not complain.


supermacro:
Since CI seems to be stuck on this PR for some strange reason, I just ran the `lint` check on my own machine. I got lint errors when I removed these directives. As such I will close this PR.

Issue #203, opened 2020-12-07
Add Synchronous API to wrap a synchronous function that throws


Liam-Tait:
Synchronous functions that throw are quite common.

I suggest a function is added to the Synchronous API to wrap these functions to return `ok` or `err`.

This could be achieved using a higher order function to wrap the synchronous function that throws

A built in example of this is `JSON.stringify`

The following error throws `TypeError: Converting circular structure to JSON`
```ts
const circularReference =  {};
circularReference.myself = circularReference;
JSON.stringify(circularReference);
```

I would suggest the utility function would wrap the function keeping the same arguments and modifying the return type to be
a Result

```ts
const JsonStringify = wrapSynchronousFn(JSON.stringify);
const jsonResult = JsonStringify(circularReference);
JsonStringify(circularReference).isOk(); // false
```

An initial pass could look something like the following. I don't think it would be possible to infer the thrown Errors, so this may need to be modified

```ts
function wrapSynchronousFn<ARGUMENTS extends readonly any[], ReturnValue, Fn extends (...args: ARGUMENTS) => ReturnValue, E>(
  fn: Fn
): (...args: ARGUMENTS) => Result<ReturnValue, E> {
  return (...args: ARGUMENTS) => {
    try {
      const result = fn(...args);
      return ok(result);
    } catch (e) {
      return err(e);
    }
  };
}
```

Please let me know if this direction would be accepted. I am happy to work on this with any suggested changes.

paduc:
Hi @Liam-Tait, I could see myself enjoying such a helper !

I would however suggest a syntax closer to `ResultAsync.fromPromise`, maybe `Result.fromThrowable`.

I would use it like this:
```typescript
const res = Result.fromThrowable( () => fnThatMayThrow(), (e) => new MyCustomError(e.message) );
// res is Result<ReturnType<typeof fnThatMayThrow>, MyCustomError>
```

Liam-Tait:
@paduc Thanks for taking the time to review this suggestion.


As `Result` is a type and not a class. this would have to be a function to work.

Adding an optional errorFn makes sense and keeps the API consistent

I suggest that having the function be higher order promotes better readability and re-usability.
In the wiki there is an example for Wrapping 3rd party code to localize exceptions
https://github.com/supermacro/neverthrow/wiki/Error-Handling-Best-Practices#synchronous-example
Which lends itself to the wrapping approach.

With above changes this would become either

**Wrapping**
```ts
[...]

PR #204, opened 2020-12-08
Add fromThrowable utility function

Liam-Tait:
It is common within javascript libraries to throw within a synchronous function

This adds a utility function fromThowable that wraps a function and creates a
function with the same signature except instead of throwing, a Result will be
returned. Ok if the original function succeds, or Err if the function throws.

As we cannot infer the type of the thrown error, this will be of unknown type
unless a mapping function is provided.

Discussed in #203

### DIFF
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -274,6 +274,50 @@ describe('Result.Err', () => {
   })
 })
 
+describe('Result.fromThrowable', () => {
+  it('Creates a function that returns an OK result when the inner function does not throw', () => {
+    const hello = (): string => 'hello'
+    const safeHello = Result.fromThrowable(hello)
+
+    const result = hello()
+    const safeResult = safeHello()
+
+    expect(safeResult).toBeInstanceOf(Ok)
+    expect(result).toEqual(safeResult._unsafeUnwrap())
+  })
+  it('Creates a function that returns an err when the inner function throws', () => {
+    const thrower = (): string => {
+      throw new Error()
+    }
+
+    // type: () => Result<string, unknown>
+    // received types from thrower fn, no errorFn is provides therefore Err type is unknown
+    const safeThrower = Result.fromThrowable(thrower)
+    const result = safeThrower()
+
+    expect(result).toBeInstanceOf(Err)
+    expect(result._unsafeUnwrapErr()).toBeInstanceOf(Error)
+  })
+
+  it('Accepts an error handler as a second argument', () => {
+    const thrower = (): string => {
+      throw new Error()
+    }
+    type MessageObject = { message: string }
+    const toMessageObject = (): MessageObject => ({ message: 'error' })
+
+    // type: () => Result<string, MessageObject>
+    // received types from thrower fn and errorFn return type
+    const safeThrower = Result.fromThrowable(thrower, toMessageObject)
+    const result = safeThrower()
+
+    expect(result.isOk()).toBe(false)
+    expect(result.isErr()).toBe(true)
+    expect(result).toBeInstanceOf(Err)
+    expect(result._unsafeUnwrapErr()).toEqual({ message: 'error' })
+  })
+})
+
 describe('Utils', () => {
   describe('`combine`', () => {
     describe('Sync `combine`', () => {
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ For asynchronous tasks, `neverthrow` offers a `ResultAsync` class which wraps a 
     - [`Result.asyncAndThen` (method)](#resultasyncandthen-method)
     - [`Result.match` (method)](#resultmatch-method)
     - [`Result.asyncMap` (method)](#resultasyncmap-method)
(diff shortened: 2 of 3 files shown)


src/utils.ts
```diff
@@ -1,6 +1,28 @@
...
 
+/**
+ * Wraps a function with a try catch, creating a new function with the same
+ * arguments but returning `Ok` if successful, `Err` if the function throws
+ *
+ * @param fn function to wrap with ok on success or err on failure
+ * @param errorFn when an error is thrown, this will wrap the error result if provided
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function fromThrowable<Fn extends (...args: readonly any[]) => any, E>(
```
supermacro, line 12: I'm fairly confident that you can replace usage of `any` by using the `infer` keyword in this case. Give it a shot.

```suggestion
export function fromThrowable<Fn extends (...args: readonly (infer U)[]) => any, E>(
```

More info:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#type-inference-in-conditional-types

---

Secondly, you could get rid of the `any` return type by declaring a additional type argument.

```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
export function fromThrowable<T, E, Fn extends (...args: readonly (infer U)[]) => T>(
```

src/utils.ts
```diff
@@ -1,6 +1,28 @@
...
+  fn: Fn,
+  errorFn?: (e: unknown) => E,
```
supermacro, line 14: Question for you: If `fromThrowable` is called without a `errorFn` argument, then what does the tsc compiler infer `E` to be? I don't know for certain (as I haven't actually checked out the code), but I think this situation may require having conditional types. 

https://www.typescriptlang.org/docs/handbook/2/conditional-types.html

tests/index.test.ts
```diff
@@ -275,6 +275,54 @@ describe('Result.Err', () => {
...
+  describe('fromThrowable', () => {
+    it('Creates a function that returns an OK result when the inner function does not throw', () => {
+      const hello = (): string => 'hello'
+      const safeHello = fromThrowable(hello)
+
+      const result = hello()
+      const safeResult = safeHello()
+
+      expect(safeResult.isOk()).toBe(true)
+      expect(safeResult.isErr()).toBe(false)
```
supermacro, line 287: There's already a test case that ensures that if `isOk` is true then `isErr` is false, and vice versa. No need to assert both here.

tests/index.test.ts
```diff
@@ -275,6 +275,54 @@ describe('Result.Err', () => {
...
+        throw new Error()
+      }
+
+      // type: () => Result<string, unknown>
+      // received types from thrower fn, no errorFn is provides therefore Err type is unknown
+      const safeThrower = fromThrowable(thrower)
+      const result = safeThrower()
+
+      expect(result.isOk()).toBe(false)
+      expect(result.isErr()).toBe(true)
```
supermacro, line 302: Same as comment above.

supermacro (changes requested):
Looks great!

As mentioned in #203, please move this function definition inside of `result.ts` within a `Result` namespace.

tests/index.test.ts
Liam-Tait: Will remove

tests/index.test.ts
Liam-Tait: Will remove

src/utils.ts
Liam-Tait: If the errorFn is not provided, `E` will be `unknown`. Which is not amazing, but I don't think there is a way to infer the error type, so is probably the most accurate.
Another option would be to make `errorFn` required

src/utils.ts
Liam-Tait: Unfortunately trying `infer U` causes a typescript error
> `'infer' declarations are only permitted in the 'extends' clause of a conditional type.`

This can be changed to `unknown` with no adverse effect
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
export function fromThrowable<Fn extends (...args: readonly unknown[]) => any, E>(
```
---

Changing the return type by declaring an additional type argument T, makes the response OK be `unknown`

`() => Result<unknown, E>`.

src/utils.ts
supermacro: Having the error function as optional is preferable. Having an `unknown` error value by default is totally fine in my opinion. 

As long as `any` types are not leaked I'm happy!

src/utils.ts
Liam-Tait: The `any` is not leaked to the return type so I think we can go ahead with this? 
I happy to remove the `any`, but am unsure of an approach at the moment

Liam-Tait:
Updated! Please take a look.

supermacro (approved):

paduc:
Thanks @Liam-Tait ! This is awesome !

Issue #211, opened 2020-12-18
Any best practices for testing with Jest?


bluenote10:
I'm using Jest as a test framework. My naive attempts at writing tests for functions that return a `Result` typically look like:

```ts
expect(callSomeFunction("with", "some", "args")).isOk()).toEqual(true);
expect(callSomeFunction("with", "some", "args"))._unsafeUnwrap()).toEqual(someExpectation);
```

This is a bit clumsy because:
- I need to double all test calls just for the `isOk()` and `_unsafeUnwrap()` parts. Especially if the `callSomeFunction("with", "some", "args")` expression is long, this causes quite some overhead.
- When the function fails, the runner only outputs that `isOk()` wasn't `true`. Any helpful information in the error object remains hidden. I frequently copy paste the expression into a third line for an ad-hoc `console.log` just to get some information, which of course feels wrong ;).
- Using `_unsafeUnwrap` and `_unsafeUnwrapErr` basically everywhere looks a bit ugly and feels like an anti-pattern.

I'm be no means a Jest expert and still have to check if it offers something that could help here.

I just wanted to bring up the topic, perhaps there are already some better approaches?

supermacro:
> Using _unsafeUnwrap and _unsafeUnwrapErr basically everywhere looks a bit ugly and feels like an anti-pattern.

I have been considering removing these methods for a while. But now I'm realizing that they make great testing utilities. 

I think that usage of `_unsafeUnwrap` and `_unsafeUnwrapErr` in "real code" (outside of tests) is definitely an antipattern, however they make for great testing utilities. 

Here's why I think this:

- Both `_unsafeUnwrap` and `_unsafeUnwrapErr` **imply** that a value is either a `Ok` or an `Err` ... otherwise they throw
  - Right now, when these methods throw, [they don't provide info on what value they actually contain](https://github.com/supermacro/neverthrow/blob/master/src/result.ts#L83) .. they just say "woops something went wrong" - so this could be improved for sure.
  - There could also be a new method called `toJson` or something along those lines which might be useful for testing / debugging, along with serialization or other use cases. 

- By using `_unsafeUnwrap` or `_unsafeUnwrapErr` you don't need to duplicate code.

```typescript
expect(callSomeFunction("with", "some", "args"))._unsafeUnwrap()).toEqual(someExpectation);

// the above also implies that you expect callSomeFunction to return a Ok value
//
[...]

Issue #11, opened 2020-01-31
Documentation mismatch for match() 


TheFalseFriend:
Been tinkering with neverthrow - it's a nice wee library @gDelgado14.

I've come across an inconsistency between the documentation (including at [npm](https://www.npmjs.com/package/neverthrow)).

The docs state for match() the following;
```
match<U, A>(
  okFn: (t:  T) =>  U,
  errFn: (e:  E) =>  A
):  U | A => { ... }
```
This suggests we could get back an `A` or a `U` when we call match(). However the code as _currently implemented_ is written as;
```
match<U, A>(
  okFn: (t:  T) =>  A,
  errFn: (e:  E) =>  A
):  A => { ... }
```  
Therefore in actual usage we can only get back an `A`. 

My questions are thus:
* which case is correct? Presumably the ```U | A``` case.
* _if_ the ```A``` case is correct, I'd be curious to understand the specific reason for enforcing more restrictive usage, from a types perspective (I'm new to TypeScript)?
* why implement match() as a property and not a method?

Thanks in advance!

gDelgado14:
Good catch @TheFalseFriend! 

> which case is correct? Presumably the U | A case.

The documented type definition was accurate at one point. See this repo's initial commit: 

[Here](https://github.com/gDelgado14/neverthrow/commit/3518425c3884e93b5915cf0205ce58355fd0c766#diff-13618125a1b532b291c07ad8ac3dc6b8R51) and [here](https://github.com/gDelgado14/neverthrow/commit/3518425c3884e93b5915cf0205ce58355fd0c766#diff-13618125a1b532b291c07ad8ac3dc6b8R95)

I can't recall why I changed it, but I did so here:

https://github.com/gDelgado14/neverthrow/commit/a07287e49cbe6b9276466966f418743051e0192c

However, I forgot to update the docs 🙈

I am always happy to accept pull requests to fix things such as incorrect documentation!

gDelgado14:
> if the A case is correct, I'd be curious to understand the specific reason for enforcing more restrictive usage, from a types perspective (I'm new to TypeScript)?

Returning `U | A` leads to cumbersome code, especially when your `U` and your `A` don't have a related field in order for the TypeScript compiler to discern if your value is indeed an `A` or a `U`. 

For more info on what I mean, check out the TS docs on [discriminated unions](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions).

[...]

PR #214, opened 2020-12-22
Make ok and err comparable

bluenote10:
Looking into #211 a bit more closely, I noticed that it is really only the `match` function that for some reason wasn't implemented as a function attached to the prototype, but rather as a property. This not only has an unnecessary memory overhead, but was also the reason why `Ok` and `Err` instances weren't comparable. I'm not sure why it has been done like that, but at least changing it does not break any test and would enable deep comparison.

Also, in the case of an coincidental `expect(ok(42)).toEqual(err(42))` Jest actually prints something quite meaningful:

image

So they cannot accidentally match because of the different property name. On first glance this would basically solve #211 for me.

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -71,7 +71,7 @@ export class Ok<T, E> {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  match = <A>(ok: (t: T) => A, _err: (e: E) => A): A => {
+  match<A>(ok: (t: T) => A, _err: (e: E) => A): A {
     return ok(this.value)
   }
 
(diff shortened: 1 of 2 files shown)


TheFalseFriend:
To chime in @bluenote10 - see [here](https://github.com/supermacro/neverthrow/issues/11#issuecomment-582059639) c.f. `match` being a property. Based on that thread, sounds like there isn't a technical issue at play (sans compiler errors at the time) so you're probably good to change to a function on the prototype - I defer to @supermacro to confirm for sure.

supermacro:
Ohhh interesting, I hadn't connected the dots that having the `match` function implemented as a prototype method would then allow all instances to share the same reference!

Yup, I'm totally fine with this change. 

Thanks @TheFalseFriend for finding the context!

supermacro (approved):

PR #223, opened 2021-01-08
fix ResultAsync.andThen link

dferber90:
At the moment, in the table of contents, the link of `ResultAsync.andThen` points to the `Result.asyncAndThen` section by accident. This fixes the link.


supermacro (approved):

supermacro:
Hmm, interesting that both headings produce basically the same anchor link.

Thanks for the contribution @dferber90 😃

PR #229, opened 2021-01-28
Export factories directly

mariosant:
Hello! This PR is intended to start a discussion for ergonomics.

I have recently discovered this great library and introduced to a colleague of mine. I felt it was a bit awkward to use the available factories, so I am raising this PR to discuss about it.

This PR makes factories available as a top level export.

It will simplify the usage overall, as it will be more straightforward to use.

```
import { fromThrowable } from 'neverthrow'

// instead of

import { Result } from 'neverthrow'

const { fromThrowable } = Result
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ For asynchronous tasks, `neverthrow` offers a `ResultAsync` class which wraps a 
     - [`ResultAsync.match` (method)](#resultasyncmatch-method)
   + [Utilities](#utilities)
     - [`combine`](#combine)
+    - [`fromThrowable`](#fromThrowable)
+    - [`fromPromise`](#fromPromise)
+    - [`fromSafePromise`](#fromSafePromise)
   + [Testing](#testing)
 * [A note on the Package Name](#a-note-on-the-package-name)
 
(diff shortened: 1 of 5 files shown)


supermacro:
Hey @mariosant, thanks for opening this PR!

I agree with the proposal overall. However, I do not agree that existing methods should be moved / deprecated.

Instead what I ask is that you re-export these functions so that folks can chose how they interact with `neverthrow`. Some folks (like myself) can keep doing:

```typescript
import { Result } from 'neverthrow'

// ...

Result.fromThrowable(blah)
```

While others (such as yourself) can import `fromThrowable` directly from `neverthrow` as you demonstrated above. 

This, of course, also implies updating the `README` so that people are aware if their options.

Let me know what you think 🙂

mariosant:
I think it's a great suggestion, @supermacro ! I will ping you when the PR gets updated.

mariosant:
Hey @supermacro, I have updated this. Please take a look when possible!

mariosant:
Subtle ping for @supermacro in case you missed the notification.

supermacro (approved):

supermacro:
Hey @mariosant thanks for the contribution! Since there was a new `beta` [release recently](https://github.com/supermacro/neverthrow/releases/tag/v4.1.0-beta.0) I have to decide when this upstream work will land in the npm default tag.

What I am thinking is that I create a new beta release under `v4.1.0-beta.0` for now. I still don't yet feel comfortable moving `v4.1.0` into the default npm tag.

Expect this new beta release to land sometime this week.

supermacro:
For future reference (and to anyone else reading this), I don't really pay attention to issues / pull requests on weekends 🙂 

I also just started a new job today, so there may be longer delays on PRs / issues from now on.

mariosant:
Thank you for the update, @supermacro! I don't mind delays, thought you might have missed the notification or something.
Good luck with your new role!

Issue #187, opened 2020-11-20
Add function docs to help users who are using intellisense or other language servers for inline documentation.


supermacro:
Context: https://github.com/supermacro/neverthrow/issues/38#issuecomment-658385438

Liam-Tait:
I'm happy to do some work on this given a direction.

I suggest two parts are added to each inline comment

1. A short description of what the unit of code does
2. `@param` description for each parameter

I don't think this needs to be as detailed or exactly the same as the README. Especially since this would double documentation maintenance by having two sources of truth.  Clear and concise descriptions should do

Suggest this is split into a pr for each code unit and progressively add these docs

supermacro:
Hey @Liam-Tait, yeah that works great. As long as [tsdoc](https://github.com/Microsoft/tsdoc) conventions are followed, I'm happy!

One thing I'm just realizing may be a challenge is the fact that methods are implemented twice for the `Result` type (once for `Ok` and again for `Err`), so just something to keep in mind and see how it works with intellisense features.

Liam-Tait:
This took awhile for me to get back to. Happy to be assigned this task.

I have created a draft pr @supermacro

supermacro:
Improvement is now live: https://github.com/supermacro/neverthrow/releases/tag/v4.2.1

Note that `ResultAsync` and the utils functions are not yet abiding by this format, but this is a great start. I'll eventually get around to update `ResultAsync` and `utils`.

PR #248, opened 2021-03-09
Add inline function docs for Ok and Err

Liam-Tait:
Based on discussion in #187 

The descriptions are trimmed down versions from the documentation at this stage.

I have created an interface `IResult`, by extending this interface the inline function docs are propagated to both `Ok` and `Error`

Keen to get specifically get some feedback on the descriptions and/or the approach of extending a shared interface

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -31,7 +31,129 @@ export const ok = <T, E>(value: T): Ok<T, E> => new Ok(value)
 
 export const err = <T, E>(err: E): Err<T, E> => new Err(err)
 
-export class Ok<T, E> {
+interface IResult<T, E> {
+  /**
+   * Used to check if a `Result` is an `OK`
+   *
+   * @returns `true` if the result is an `OK` variant of Result
+   */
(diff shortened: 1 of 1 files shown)


supermacro:
This is great! I hadn't considered having a shared interface so as to have a single place to write function docs. 

I'm happy with both the descriptions and the usage of a shared abstract interface. 

If we adhere to tsdoc syntax, then it may be possible to generate the README docs based off of these docs, thus eliminating the duplication of documentation... but this is for another day.

Liam-Tait:
Sweet, thank you @supermacro, would you like all the function docs in one PR, or create several smaller sections so it can be done bit by bit?

supermacro:
Bit by bit is fine. This is a good first start. 

FYI, I transitioned CI over to circle, so you'll need to rebase onto the latest commit on this project before I can approve + merge.

Liam-Tait:
@supermacro I have rebased this, thanks

supermacro (approved):

Issue #197, opened 2020-11-29
Convert Array of Result<T,E> to Result<T,E[]>


afontcu:
Hi! First of all, thanks for this library. I'm starting to learn how to deal with it, and found a situation I'm not sure how to handle. 

Use case: I want to validate a UI field, and print **all** failed validations. So, I think I need to convert an array of Results to a result with an array of errors.

```ts
const validateMinLength = (s: string) =>
  s.length > 8 ? ok(s) : err('too short')

const validateStrength = (s: string, getStrength: (s: string) => number) =>
  getStrength(s) > 25 ? ok(s) : err('too weak')
```

What's the best way to combine those results, and either return `s` if everything is fine or `['too short', 'too weak']` if both validations failed (so, without short-circuiting)?

Thanks! 🙌

supermacro:
Hey @afontcu, 

If I understand correctly, you have a UI field that contains some data represented by a type `T` and that field contains one or more validation rules, which I'll represent by a type `E[]`.

Having an initial `Array<Result<T, E>>` doesn't make sense to me because you're saying, "I may have one or more `T` values for my single UI field".

So what I would think you need is something like this:

```typescript
type ErrorMessage = string

type Validator<T> = (val: T) => ErrorMessage | null

const validateMinLength = (s: string):  =>
  s.length <= 8 ? 'too short' : null

const validateStrength = (getStrength: (s: string) => number) => (s: string): ErrorMessage | null =>
  getStrength(s) <= 25 ? 'too weak' : null

const validateMyField = <T>(validators: Validator<T>, val: T): ErrorMessage[] =>
  validators.reduce(
    (errors, validator) =>
      const maybeError = validator(val)
      
     if (maybeError) {
        return [...errors, maybeError]
     }
    
     return errors
  , [])
```

With this you can then have logic that takes a UI field value of type `T` and returns a `Result<T, E[]>`

afontcu:
Sounds good :) I guess I was trying to "fit" my idea into Result, but it wasn't meant for that.

Thank you for your time! Appreciated.

PR #251, opened 2021-03-11
Implement combineWithAllErrors

tobloef:
This PR implements a utility function `combineWithAllErrors`, which I needed for a project. It basically does the same as `combine`, but instead of returning `Result<[T1, T2, ...], E1 | E2 | ...>` it returns `Result<[T1, T2, ...], (E1 | E2 | ...)[]>`. That is, instead of short-circuiting like the normal `combine`, it returns a list of all the errors it encountered, if the result is an error. It's only for the synchronous `Result` right now, not `ResultAsync`.

This may or may not be something you want to include in the library, but I found it to be very useful, and I believe it might solve #197 as well. We can change the name (or any other aspects) if you wish, I just thought that I would share the changes I made.


supermacro:
Hey, @tobloef this is a great idea. 

A few things:

- I just switched over from TravisCI to CircleCI, so I need you to please rebase your changes to the latest commit in this project. 
- If I'm not mistaken / making any wrong assumptions, getting this to work with `ResultAsync` is trivial and only requires a `Promise.all` and `fromSafePromise` call
 - see `combine` for `ResultAsync` (https://github.com/supermacro/neverthrow/blob/master/src/utils.ts#L57)  - Could you please update the docs (specifically the [Utilities Section](https://github.com/supermacro/neverthrow#utilities)) to include this new function. A short example would be great too.

tobloef:
Hey @supermacro, I've updated the README, implemented the async version and fixed the stuff the linter was complaining about. Let me know what you think.

supermacro (approved):
Looks good to me! Thank you for the contribution 🙂

PR #311, opened 2021-07-01
Fixed #300: Updated fromThrowable type signature to allow inner function to have arguments.

### LINKED ISSUE #300: fromThrowable type signature too restrictive?
Hello!

I think I've found a bug in neverthrow 4.2.1.

With this code, more or less lifted from the [documentation](https://github.com/supermacro/neverthrow#resultfromthrowable-static-class-method):

```ts
import { Result } from "neverthrow";

const foo = Result.fromThrowable(JSON.parse, () => {});
```

... I get this transpilation failure:

```
Argument of type '(text: string, reviver?: ((this: any, key: string, value: any) => any) | undefined) => any' is not assignable to parameter of type '(...args: readonly unknown[]) => any'.
  Types of parameters 'text' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'string'.ts(2345)
```

I haven't really dug in yet, but at first blush it seems like `fromThrowable` might be too opinionated about the arguments of the wrapped function.

It does seem to work to change the type signature of `fromThrowable` from:
`<Fn extends (...args: readonly unknown[]) => any, E>`
... to...
`<Fn extends (...args: readonly any[]) => any, E>`

Any thoughts?

universalhandle:
Fixes #300

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -11,7 +11,7 @@ export namespace Result {
    * @param errorFn when an error is thrown, this will wrap the error result if provided
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  export function fromThrowable<Fn extends (...args: readonly unknown[]) => any, E>(
+  export function fromThrowable<Fn extends (...args: readonly any[]) => any, E>(
     fn: Fn,
     errorFn?: (e: unknown) => E,
   ): (...args: Parameters<Fn>) => Result<ReturnType<Fn>, E> {
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -315,6 +315,19 @@ describe('Result.fromThrowable', () => {
     expect(safeResult).toBeInstanceOf(Ok)
     expect(result).toEqual(safeResult._unsafeUnwrap())
   })
+
+  // Added for issue #300 -- the test here is not so much that expectations are met as that the test compiles.
+  it('Accepts an inner function which takes arguments', () => {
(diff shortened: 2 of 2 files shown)


supermacro (approved):
LGTM 👍 interesting that changing the type from `unknown` to `any` fixed the issue. Makes sense in retrospect though.

universalhandle:
Hmm, if it makes sense to you, I'd love to hear your thoughts on it.

The error message is weird: `Type 'unknown' is not assignable to type 'string'`. My best guess as to why this doesn't work is that our definition of `Fn` says that `args` is an array of `unknown`s, but the passed function (`JSON.parse`) expects a string. `fromThrowable` doesn't do any type narrowing on `args` before passing it to `fn` (i.e., `JSON.parse`), which is the only way I can see we're trying to assign `unknown` to `string`. (Of course `fromThrowable` can't and shouldn't try to do any type narrowing, because we can't know what sort of function is being passed to it.)

supermacro:
My thinking / understanding: 

You cannot assign anything that is of type `unknown` to a variable that is more specific, such as `string`. Conversely, you are allowed to assign something of type `any` to a variable that is more specific.

By using `any` we are able to allow for different degrees of arity (number of args) with varying arg types for our generic type `Fn`. When we pass a function into `fromThrowable`, then `Fn` takes on the definition of the function we passed in and it adheres to the "looser" constraint of `any[]` for args.

PR #349, opened 2021-10-07
Improving the inference of `andThen` and `orElse` methods

sidhu663:
Allows for improved inference when returning more complex types in the `andThen` and `orElse` functions without needing to explicitly specify the types.

The premise of the change is two-fold:
1. Defaulting `err` and `ok` to `never` instead of letting them become unknown. A type union-ed with `never` is the original type (`SomeType | never === SomeType`) so defaulting the constructor type parameters to never prevents the introduction of `unknown`. Since a type union-ed with another type is unknown (`SomeType | unknown === unknown`), avoiding unknowns in the constructor prevents loss of type information.

2. Adding overloads to the methods where the generic parameter is the full `Result` or `ResultAsync`. This allows the use of conditional inference in typescript (`type UnwrapArray<T> = T extends Array<infer U> ? U : unknown`) while still allowing for a fallback where you can provide the args for the new ok / err types

Examples:

```typescript
    // before: Result<unknown, unknown>, after: Result<number, string>
    const example1 = ok(123).andThen(t => {
        if (t > 100) {
            return ok(t)
        } else {
            return err('failure')
        }
    })

    // before: Result<unknown, unknown>, after: Result<number, never>
    const example2 = ok(123).andThen(t => {
        if (t > 100) {
            return ok(t)
        } else {
            return ok(0)
        }
    })

    // before: Result<unkown, unknown>, after: Result<Response, string | Response>
    const example3 = err<Response, string>('unexpected error').andThen(t => {
        if (t.ok) {
            return ok(t)
        } else {
            return err(t)
        }
    })

    // before: Argument of type '(t: Response) => Ok<string, unknown> | Ok<Response, unknown>' is not assignable to parameter of type '(t: Response) => Result<string, unknown>'.
    // after: Result<string | Response, string>
    const example4 = err<Response, string>('unexpected error').andThen(t => {
        if (t.ok) {
            return ok(t)
        } else {
            return ok('200')
        }
    })

   // before: Result<Response, unkown>
    // after: Result<Response, number>
    const example5 = errAsync<Response, string>('unexpected error').orElse(e => {
        if (e === 'unexpected') {
            return ok(new Response())
        } else {
            return err(500)
        }
    })
```

### DIFF
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -47,7 +47,7 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
   }
 
 
-    andThen<R extends Result<unknown, unknown>>(
+  andThen<R extends Result<unknown, unknown>>(
     f: (t: T) => R,
   ): ResultAsync<InferOkTypes<R>, InferErrTypes<R> | E>
   andThen<R extends ResultAsync<unknown, unknown>>(


sidhu663:
Sounds good, I'll add some more tests and examples. As for your example, it should be `Result<string, string | boolean>` correct? Since the previous error type is unioned with the new error type. (And it is inferred automatically in my code!)

PR #377, opened 2022-01-15
Fix misleading inaccuracy in `match` documentation

tam-carre:
The `Result` `match` documentation incorrectly claims that `match` is in some way equivalent to chaining `map` and `mapErr`. While this is true if you only perform side effects (as this PR clarifies), this is misleading as `match` returns an unwrapped value while chaining `map` and `mapErr` returns a `Result`.

It would, however, be accurate to say that chaining `map` and `unwrapOr` is equivalent to calling `match` in cases where the error parameter is unused.

This PR clarifies this and provides examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,31 +449,52 @@ class Result<T, E> {
 ```
 
 `match` is like chaining `map` and `mapErr`, with the distinction that with `match` both functions must have the same return type.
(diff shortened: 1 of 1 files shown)


supermacro (approved):
Thank you :)

Issue #381, opened 2022-02-05
Bug: overloaded combine API is unsafe


evelant:
The `combine` API was crashing on me in production, turns out there is a corner case due to typescript overloading

```typescript
const res: ResultAsync<void, void>[] = [errAsync(undefined)]
const res2: ResultAsync<void, void>[] = [okAsync(undefined)]
const combined = combine([combine(res), combine(res2)]).then((res) => {
    //no problem
})
```

```typescript
const res: ResultAsync<void, void>[] = []
const res2: ResultAsync<void, void>[] = [okAsync(undefined)]

//crash! TypeError: result.isErr is not a function
const combined = combine([combine(res), combine(res2)]).then((res) => {
   
})
```

The overload here is the issue https://github.com/supermacro/neverthrow/blob/524a15a100034f2d74a2a39e1fb1ed5937da41a2/src/utils.ts#L65-L71

If you ever pass an empty array to `combine` you get unexpected behavior since it _always_ returns a `Result` in that case even if the type should be a `ResultAsync` which results in a crash in some cases.

Unfortunately I don't see any way to resolve this other than separating `combine` into `combine` and `combineAsync`. Thoughts?

supermacro:
Hey, good find! I'm surprised this issue hasn't come up sooner.

> Unfortunately I don't see any way to resolve this other than separating combine into combine and combineAsync. Thoughts?

Hmm, why couldn't we do something like this:

```typescript
export function combineWithAllErrors(list: any): any {
  if (list.length === 0) return []

  if (list[0] instanceof ResultAsync) {
    return combineResultAsyncListWithAllErrors(list)
  } else {
    return combineResultListWithAllErrors(list)
  }
}
```

Intuitively this seems to be a good solution to me, but maybe I'm missing something?

supermacro:
Wait what am I saying my solution doesn't even compile 🤣 

Yeah I get what you're saying now. There's no way to know if we should return a `ResultAsync` or a `Result` when there's an empty list.

I think another solution in contrast to what you proposed is to define a type that is not `Result<unknown, unknown>[]` or `T extends readonly ResultAsync<unknown, unknown>[]`

We'd want something that is seen as different than the array type `[]`

Perhaps non-empty array?

```
type NonEmptyArray<T> = [ T, ...T[] ]
```

Here's an example that intentionally does not compile:

```typescript

type NonEmptyArray<T> = [ T, ...T[] ]

const maybeEmpty: string[] = []

[...]

PR #405, opened 2022-07-26
Feat/split combine into sync and async ii

tam-carre:
Resolves https://github.com/supermacro/neverthrow/issues/381

Introduces breaking API changes

This PR removes the `combine` and `combineWithAllErrors` exports and creates `Result.combine`, `Result.combineWithAllErrors`, `ResultAsync.combine`, `ResultAsync.combineWithAllErrors`.

The "Utilities" section of the README has been removed as it would become just references to Result and ResultAsync sections.


tam-carre:
Not the right file versions, hold on =\

Issue #226, opened 2021-01-15
Unbounded lists of results without a type annotation don't have enough type information


wiktor-obrebski:
This is next proposition to solve problem with #224.

I feel I learned a lot about TypeScript in last week ;) Today I tried again to provide proper types for Heterogeneous `combine` function.

First, I would like to provide my new proposition, implementation of `combine` types, based on our findings in #224 and my experimetns.

```ts
import { Result, Ok, Err } from 'neverthrow';

declare function combine<T extends readonly Result<unknown, unknown>[]>(
  results: [...T]
): Result<ExtractOkTypes<T>, ExtractErrTypes<T>[number]>;

// Given a list of Results, this extracts all the different `T` types from that list
type ExtractOkTypes<T extends readonly Result<unknown, unknown>[]> = {
  [Key in keyof T]: T[Key] extends Result<unknown, unknown>
    ? ExtractOkFromUnion<T[Key]>
    : never;
};

// Given a list of Results, this extracts all the different `E` types from that list
type ExtractErrTypes<T extends readonly Result<unknown, unknown>[]> = {
  [Key in keyof T]: T[Key] extends Result<unknown, unknown>
    ? ExtractErrFromUnion<T[Key]>
    : never;
};

// need to be separated generic type to run it for every element of union T separately
type ExtractOkFromUnion<T extends Result<unknown, unknown>> = T extends Ok<
  infer V,
  unknown
> // filter out "unknown" values
  ? V extends {}
    ? V
    : never
  : never;

// need to be separated generic type to run it for every element of union T separately
type ExtractErrFromUnion<T extends Result<unknown, unknown>> = T extends Err<
  unknown,
  infer E
> // filter out "unknown" values
  ? E extends {}
    ? E
    : never
  : never;

```

wiktor-obrebski:
And there is a pack of types tests where I tried to cover most of the cases we discussed before.

```ts
import { ok, err, Result, Ok, Err } from 'neverthrow';

// definition of atomic data
class InvalidTest {}

const ok1 = ok('my value1');
const ok2 = ok(2);
const err1 = err(new InvalidTest());

// definition of many kinds of result-related collections that can occur
const arr = [ok1, ok2];
const arrWithErr = [ok1, err1];

const arrOfRes: (Result<string, unknown> | Result<number, unknown>)[] = [
  ok1,
  ok2,
];
const arrayOfResWithErr: (
  | Result<string, unknown>
  | Result<unknown, InvalidTest>
[...]

PR #418, opened 2022-09-28
Provides the types for `combine` and `combineWithAllErrors`, solves #226

incetarik:
Hello,

This PR solves the type problem of the `combine` and `combineWithAllErrors` methods as described [here](https://github.com/supermacro/neverthrow/issues/226).

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -308,149 +308,160 @@ export const fromThrowable = Result.fromThrowable
 
 // Given a union, this extracts the intersection of the unions by casting it
 // to a parameter of a function and then inferring from it.
-type IntersectOf<U extends any>
-  = (U extends unknown ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+type IntersectOf<U extends unknown> = (U extends unknown ? (k: U) => void : never) extends (
+  k: infer I,
+) => void
+  ? I
+  : never
 
 // Given a union, this extracts the last possible value.
-type Last<U extends any>
-  = IntersectOf<U extends unknown ? (x: U) => void : never> extends (x: infer P) => void ? P : never
+type Last<U extends unknown> = IntersectOf<U extends unknown ? (x: U) => void : never> extends (
+  x: infer P,
+) => void
+  ? P
+  : never
 
 /// Given a union and a member, this excludes the member.
-type ExcludeUnionItem<U extends any, M extends any> = U extends M ? never : U
+type ExcludeUnionItem<U extends unknown, M extends unknown> = U extends M ? never : U
 
 // Given two types, this checks whether the A1 is extending the A2, including
 // the `never` type in calculation. If A1 is `never`, then this will be `0`.
-type IsExtending<A1 extends any, A2 extends any>
-  = [ A1 ] extends [ never ] ? 0 : A1 extends A2 ? 1 : 0
+type IsExtending<A1 extends unknown, A2 extends unknown> = [A1] extends [never]
+  ? 0
+  : A1 extends A2
+  ? 1
+  : 0
 
 // Given an array and a type, this extends the array by appending the `A`.
-type Append<L extends ReadonlyArray<any>, A extends any> = [
-  ...L,
-  A,
-]
+type Append<L extends ReadonlyArray<unknown>, A extends unknown> = [...L, A]
 
 // Given a union, this gives the array of the union members.
-type MemberListOf<
-  Union,
-  MemberList extends ReadonlyArray<any> = [],
-  LastU = Last<Union>
-  > = {
-    0: MemberListOf<ExcludeUnionItem<Union, LastU>, Append<MemberList, LastU>>
-    1: MemberList
-  }[ IsExtending<[ Union ], [ never ]> ]
+type MemberListOf<Union, MemberList extends ReadonlyArray<unknown> = [], LastU = Last<Union>> = {
+  0: MemberListOf<ExcludeUnionItem<Union, LastU>, Append<MemberList, LastU>>
+  1: MemberList
+}[IsExtending<[Union], [never]>]
 
 // This is a helper type to prevent infinite recursion in typing rules.
 //
 // Use this with your `depth` variable in your types.
 type Prev = [
-  never, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
-  11, 12, 13, 14, 15, 16, 17, 18, 19, 20, ...0[]
+  never,
+  0,
+  1,
+  2,
+  3,
+  4,
+  5,
+  6,
+  7,
+  8,
+  9,
+  10,
+  11,
+  12,
+  13,
+  14,
+  15,
+  16,
+  17,
+  18,
+  19,
+  20,
+  ...0[]
 ]
 
 // Combines the both sides of the array of the results into a tuple of the
 // union of the ok types and the union of the err types.
 //
 // T     - The array of the results
-// Oks   - The initial `Ok` type values 
+// Oks   - The initial `Ok` type values
 // Errs  - The initial `Err` type values
 // Depth - The maximum depth.
-type Combine<
-  T,
-  Oks extends any[] = [],
-  Errs extends any[] = [],
-  Depth extends number = 5
-  >
-  =
+type Combine<T, Oks extends unknown[] = [], Errs extends unknown[] = [], Depth extends number = 5> =
   // If we have reached to the maximum depth, return unknown for both sides
-  [ Depth ] extends [ never ] ? Result<unknown, unknown>
-
-  // Otherwise test whether T is the list of possible values
-  : T extends [ infer H, ...infer Rest ] ? (
-    // And test whether the head of the list is a result
-    H extends Result<infer L, infer R> ? (
-      // Continue combining...
-      Combine<
-        // the rest of the elements
-        Rest,
-
-        // excluding the `unknown` type variables for `Ok` results
-        [ unknown ] extends [ L ] ? Oks : [ L, ...Oks ],
-
-        // excluding the `unknown` type variables for `Err` results
-        [ unknown ] extends [ R ] ? Errs : [ R, ...Errs ],
-
-        // and one less of the current depth
-        Prev[ Depth ]
-      >
-    ) : never // Impossible
-  ) : [ Oks, Errs ] // the results combined
+  [Depth] extends [never]
+    ? Result<unknown, unknown>
+    : // Otherwise test whether T is the list of possible values
+    T extends [infer H, ...infer Rest]
+    ? // And test whether the head of the list is a result
+      H extends Result<infer L, infer R>
+      ? // Continue combining...
+        Combine<
+          // the rest of the elements
+          Rest,
+          // excluding the `unknown` type variables for `Ok` results
+          [unknown] extends [L] ? Oks : [L, ...Oks],
+          // excluding the `unknown` type variables for `Err` results
+          [unknown] extends [R] ? Errs : [R, ...Errs],
+          // and one less of the current depth
+          Prev[Depth]
+        >
+      : never // Impossible
+    : [Oks, Errs] // the results combined
 
 // Reverses the given array
-type Reverse<A>
-  = A extends [ infer H, ...infer R ] ? (
-    [ ...Reverse<R>, H ]
-  ) : A
+type Reverse<A> = A extends [infer H, ...infer R] ? [...Reverse<R>, H] : A
 
 // Deduplicates the result, as the result type is a union of Err and Ok types.
-type Dedup<T>
-  = T extends Result<infer RL, infer RR>
-  ? Ok<RL, RR> : T
+type Dedup<T> = T extends Result<infer RL, infer RR> ? Ok<RL, RR> : T
 
 // Converts an empty array to never.
-type EmptyArrayToNever<T>
-  = T extends [] ? never : T
+type EmptyArrayToNever<T> = T extends [] ? never : T
 
 // Gets the member type of the array or never.
-type MembersToUnion<T>
-  = T extends any[] ? T[ number ] : never
+type MembersToUnion<T> = T extends unknown[] ? T[number] : never
 
 // Checks if the given type is a literal array.
-type IsLiteralArray<T>
-  = T extends { length: infer L } ? (
-    L extends number ? (
-      number extends L ? 0 : 1
-    ) : 0
-  ) : 0
+type IsLiteralArray<T> = T extends { length: infer L }
+  ? L extends number
+    ? number extends L
+      ? 0
+      : 1
+    : 0
+  : 0
 
 // Traverses an array of results and returns a single result containing
 // the oks and errs union-ed/combined.
 type Traverse<
   T,
-  Oks extends any[] = [],
-  Errs extends any[] = [],
+  Oks extends unknown[] = [],
+  Errs extends unknown[] = [],
   Depth extends number = 5
-  >
-  = Combine<T, Oks, Errs, Depth> extends [ infer Oks, infer Errs ] ? (
-    Result<EmptyArrayToNever<Oks>, Errs extends any[] ? Errs[ number ] : never>
-  ) : never
+> = Combine<T, Oks, Errs, Depth> extends [infer Oks, infer Errs]
+  ? Result<EmptyArrayToNever<Oks>, Errs extends unknown[] ? Errs[number] : never>
+  : never
 
 // Traverses an array of results and returns a single result containing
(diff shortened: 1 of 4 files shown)


incetarik:
Just realized that there are tests depending on it, I'm now improving the types and will push another commit for each test case.

incetarik:
Test are now completely passing, I'm working on linting now and then will push the new version.

supermacro:
Amazing @incetarik .. currently there's a lot on my plate at work so I will aim to have provided a thorough review within a week's time - apologies for being a bit slow / unresponsive.

tests/typecheck-tests.ts
```diff
@@ -542,7 +542,72 @@ import { ok, err, okAsync, errAsync, Result, ResultAsync } from '../src'
 
 
 (function describe(_ = 'Combine on Unbounded lists') {
-  // TODO:
-  // https://github.com/supermacro/neverthrow/issues/226
+  (function describe(_ = 'combine') {
+    (function it(_ = 'combines different results into one') {
+      type Expectation = Result<[ number, string ], Error | string[]>;
```
supermacro, line 547: This expectation is not correct - perhaps there may some miscommunication / misunderstanding of the desired goal of #226. 

The desired outcome when fixing #226 is that all the `combine` functions can correctly produce a **tuple** whose length / size is equal to the length / size of the array passed into the `combine` functions.

In other words, we need to treat the list argument to the `combine` functions as a tuple instead of a variable-length list.

Example:

When calling:

```typescript
Result.combine([
  ok(1),
  ok('string'),
  err([ 'string', 'string2' ]),
  err(new Error('error content')),
])
```

The `resultList` argument should understand that we have a fixed-length tuple whose type is:

```typescript
[ Result<number, never>, Result<string, never>, Result<never, string>, Result<never, Error> ]
```

Therefore, the `Expectation` of this test would be that we should have:

```typescript
Result<[ number, string, never, never ], Error | string[] | never>;
```

---

Additionally - in practice, `combine` would not be called with the `ok` and `err` constructors (or the `okAsync` / `errAsync` constructors), rather, we'd have a list of `Result<T, E>` values where the `T`s and the `E`s are not `unknown`.

So to make these tests a bit more realistic, I would provide explicit type arguments to the `ok` and `err` function invocations like this:

```typescript
type Expectation = Result<[ number, string, boolean ], Error | string[]>;

const result: Expectation = Result.combine([
  ok<number, string>(1),
  ok<string, string>('string'),
  err<boolean, string[]>([ 'string', 'string2' ]),
  err<boolean, Error>(new Error('error content')),
])
```

supermacro (changes requested):

tests/typecheck-tests.ts
incetarik: Ok as the types did already pass the tests in the examples that I have not added, I can still inspect the ones that you mention. Add your expectations and I'll see them. After this point, obviously, it will not be hard to do the rest. @supermacro

tests/typecheck-tests.ts
```diff
@@ -542,7 +542,72 @@ import { ok, err, okAsync, errAsync, Result, ResultAsync } from '../src'
...
+        ok('string'),
+      ]);
+    });
+
+    (function it(_ = 'combines only err results into one') {
+      type Expectation = Result<never, (string | number)[]>;
+
+      const result: Expectation = Result.combineWithAllErrors([
+        err(1),
+        err('string'),
```
supermacro, line 608: :+1:

tests/typecheck-tests.ts
```diff
@@ -542,7 +542,72 @@ import { ok, err, okAsync, errAsync, Result, ResultAsync } from '../src'
...
+
+  (function describe(_ = 'combineWithAllErrors') {
+    (function it(_ = 'combines different results into one') {
+      type Expectation = Result<[ number, string ], (Error | string[])[]>;
+
+      const result: Expectation = Result.combineWithAllErrors([
+        ok(1),
+        ok('string'),
+        err([ 'string', 'string2' ]),
+        err(new Error('error content')),
```
supermacro, line 590: Shouldn't the expectation here be:

 ```typescript
Result<[ number, string, never, never ], (Error | string[])[]>```

??

tests/typecheck-tests.ts
```diff
@@ -542,7 +542,72 @@ import { ok, err, okAsync, errAsync, Result, ResultAsync } from '../src'
...
+      ]);
+    });
+
+    (function it(_ = 'combines only err results into one') {
+      type Expectation = Result<never, number | string>;
+
+      const result: Expectation = Result.combine([
+        err(1),
+        err('string'),
+      ]);
```
supermacro, line 572:
```suggestion
      type Expectation = Result<[ never, never ], number | string>;

      const result: Expectation = Result.combine([
        err(1),
        err('string'),
      ]);
```

supermacro (changes requested):

src/result.ts
```diff
@@ -308,3 +303,165 @@ export class Err<T, E> implements IResult<T, E> {
...
+  13,
+  14,
+  15,
+  16,
+  17,
+  18,
+  19,
+  20,
+  ...0[]
+]
```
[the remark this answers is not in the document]
incetarik: Yes, that's right. AFAIK, TypeScript already does recursion for only fixed amount of times and it is why we have 20 there.

tests/typecheck-tests.ts
incetarik: I have not considered `never` types to be in order, therefore not processed them. When I'm free, I'll go back and add those and it will work. The hardest part is already done as you see. The simplest thing would be taking this result and the parameter array (which is readonly) and until their lengths matches, adding `never`s to array is the first solution I have in my mind but as I said, I'll see this when I'm free. Should not be a big problem as we could done it until this point.

src/result.ts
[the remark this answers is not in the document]
supermacro: >  TypeScript already does recursion for only fixed amount of times

If TS already has a recursion limit, why would we then need to impose our own limit?

tests/typecheck-tests.ts
```diff
@@ -542,7 +542,72 @@ import { ok, err, okAsync, errAsync, Result, ResultAsync } from '../src'
...
+      type Expectation = Result<[ number, string ], never[]>;
+
+      const result: Expectation = Result.combineWithAllErrors([
+        ok(1),
+        ok('string'),
+      ]);
+    });
+
+    (function it(_ = 'combines only err results into one') {
+      type Expectation = Result<never, (string | number)[]>;
```
supermacro, line 604: Just realized I missed this one - this should also be changed to `Result<never[], (string | number)[]>

supermacro (changes requested):

src/result.ts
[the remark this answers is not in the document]
incetarik: TS by default tries to resolve the type recursively as much as possible and that usually ends up in hanging or simply replacing the type with `never` or `unknown` as you can expect. We just make this in that way so that the TS server will stop recursing indefinitely, i.e. not showing us "Type instantiation is excessively deep and possibly infinite" error.

supermacro:
:warning:  **Potential Updates To ResultAsync functions and `utils.ts`**

I believe that the utility functions inside of `utils` (which are used by both the synchronous and asynchronous versions) might need to have their types updated. This might also mean that we no longer need to do typecasts such as the one [here](https://github.com/supermacro/neverthrow/pull/418/files#diff-fe54c0b0441ef7719a22666a2bb171f2db9e99c0ed52dd8fec96a9882e7c842dR37).

Also note that the async versions of `combine` and `combineWithAllErrors` is still using the `Extract*` utility types - which, in my mind, it seems like we no longer need (unless I am missing something). 

image

tests/typecheck-tests.ts
incetarik: Hello @supermacro, I have improved the typing system and made some of the lighter for CPU, and also easier for type matching. Yet, I have not touched anything else mentioned other than this conversation page.

tests/typecheck-tests.ts
incetarik: Updated the typing, see the commit.

tests/typecheck-tests.ts
```diff
@@ -542,7 +542,72 @@ import { ok, err, okAsync, errAsync, Result, ResultAsync } from '../src'
...
+      type Expectation = Result<never, number | string>;
+
+      const result: Expectation = Result.combine([
+        err(1),
+        err('string'),
+      ]);
+    });
+
+    (function it(_ = 'combines empty list results into one') {
+      type Expectation = Result<never, never>;
```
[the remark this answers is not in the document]
incetarik: Done

tests/typecheck-tests.ts
incetarik: Done

tests/typecheck-tests.ts
incetarik: @supermacro Could you please see the changes? Now I have updated the types as well, they are now more precise and in-order as well.

incetarik:
> ⚠️ **Potential Updates To ResultAsync functions and `utils.ts`**
> 
> I believe that the utility functions inside of `utils` (which are used by both the synchronous and asynchronous versions) might need to have their types updated. This might also mean that we no longer need to do typecasts such as the one [here](https://github.com/supermacro/neverthrow/pull/418/files#diff-fe54c0b0441ef7719a22666a2bb171f2db9e99c0ed52dd8fec96a9882e7c842dR37).
> 
> Also note that the async versions of `combine` and `combineWithAllErrors` is still using the `Extract*` utility types - which, in my mind, it seems like we no longer need (unless I am missing something).
> 
> image

I have done nothing for this. I may see this after you see the commits I have sent and approve, I think this is out of the current conversation but also something that could be done as well, right.

src/result.ts
```diff
@@ -308,3 +315,171 @@ export class Err<T, E> implements IResult<T, E> {
...
+      >
+    : never // Impossible
+  : Collected
+
+// Transposes an array
+//
+// A          - The array source
+// Transposed - The collected transposed array
+// Depth      - The maximum depth.
+type Transpose<A, Transposed extends unknown[][] = [], Depth extends number = 10> = A extends [
```
supermacro, line 379: Could we export this to have 1 or 2 type-level tests written for it, please?

tests/typecheck-tests.ts
```diff
@@ -551,6 +552,71 @@ import { ok, err, okAsync, errAsync, fromSafePromise, Result, ResultAsync } from
...
+      type Expectation = Result<[ number, string ], never[]>;
+
+      const result: Expectation = Result.combineWithAllErrors([
+        ok(1),
+        ok('string'),
+      ]);
+    });
+
+    (function it(_ = 'combines only err results into one') {
+      type Expectation = Result<never, [number, string]>;
```
supermacro, line 614:
```suggestion
      type Expectation = Result<never[], [number, string]>;
```

This is odd, it seems like there is different behaviour when passing similar arguments. Compare this expectation to this one:

https://github.com/supermacro/neverthrow/pull/418/files#diff-5e6df76f1ff5c2ddfbb0bee2e515236794778a429792a91d3b853bfecc884c16R577

supermacro (changes requested):

supermacro:
> I have done nothing for this. I may see this after you see the commits I have sent and approve, I think this is out of the current conversation but also something that could be done as well, right.

I'm happy with your approach to the async side, but we also need to apply this work to the async side:

image

the `test` variable is currently a `ResultAsync<(string | number | Date)[], string | boolean | Error>`, but it should instead be `ResultAsync<[ number, string, Date ], string | boolean | Error>` based on the work you've done in this PR.

tests/typecheck-tests.ts
incetarik: @supermacro  These two are different, one is a tuple where the other is an array. The tuple has fixed size where the array does not. I have preferred to use tuple as it would be providing a better type and length info.

src/result.ts
incetarik: @supermacro

incetarik:
> > I have done nothing for this. I may see this after you see the commits I have sent and approve, I think this is out of the current conversation but also something that could be done as well, right.
> 
> I'm happy with your approach to the synchronous side, but we also need to apply this work to the async side:
> 
> image
> 
> the `test` variable is currently a `ResultAsync<(string | number | Date)[], string | boolean | Error>`, but it should instead be `ResultAsync<[ number, string, Date ], string | boolean | Error>` based on the work you've done in this PR.

I have not done anything in the async side. I do not expect it to be anything hard as well. Yet, I have never considered them as the problem/talk was for combine and its other only, not async.

incetarik:
Good news @supermacro! The typing is also added for `ResultAsync` and it is precise as `Result`. After you see them, I have left the removals of the unused/unnecessary types to you as you may change them or reorder/update etc. Have a nice day!

supermacro:
Amazing! Thank you. I don't know if I'll have time to do a final read through of everything today, but I will try to do so by end of the week. Otherwise, I will definitely do a review by next week at the latest. And then I can send over that bounty :slightly_smiling_face:  

From a quick skim, I did notice that you added overload type definitions to the `ResultAsync.combine` and `ResultAsync.combineWithAllErrors` methods - it would be helpful to get some context on why that was done and what benefit it has.

incetarik:
> Amazing! Thank you. I don't know if I'll have time to do a final read through of everything today, but I will try to do so by end of the week. Otherwise, I will definitely do a review by next week at the latest. And then I can send over that bounty 🙂
> 
> From a quick skim, I did notice that you added overload type definitions to the `ResultAsync.combine` and `ResultAsync.combineWithAllErrors` methods - it would be helpful to get some context on why that was done and what benefit it has.

@supermacro Of course, it has a simple explanation. The TypeScript will always try to match the overloads from top to bottom. Whenever the match occurs the others are skipped. This becomes pretty useful when you want to handle the parameter for the cases where one time you pass something has narrowed, specific type such as `[1, 2, 3]` instead of `number[]`. Here you may want to do something else when you have a more accurate type. I use this to match for each element in the array. Secondly, if user passes something more generic than that, then I handle it differently and return some more general type as the return value. This is why we have overloads for these types. If you look closely, the types actually changing nothing in terms of types but the number of elements. If we are sure that there is at least one element in the array passed to the function so we have better constraints to take advantage of, otherwise we fallback to the default implementation.

tests/typecheck-tests.ts
```diff
@@ -551,6 +561,126 @@ import { ok, err, okAsync, errAsync, fromSafePromise, Result, ResultAsync } from
...
+      type Expectation = Result<[ number, string ], never>;
+
+      const result: Expectation = Result.combine([
+        ok(1),
+        ok('string'),
+      ]);
+    });
+
+    (function it(_ = 'combines only err results into one') {
+      type Expectation = Result<[ never, never ], number | string>;
```
supermacro, line 586: ### Edge case still not working as expected

When a list of `err` values is passed into any variation of `combine` (whether it's synchronous or asynchronous) **AND** there is no type argument for the success type `T` passed in (thus, the `T` becomes `never`), then we would expect to have a type akin to the one that you have written here on line 586.

**However**, when I try the code out on my machine, the inferred type does not match this expectation:

image

The Expectation is correct in this file, but for some reason the type inference logic is not returning what we expect - and for some reason the code still compiles (I don't understand how `never` can be assigned to `[ never, never ]`).

supermacro (changes requested):

supermacro:
[This](https://github.com/supermacro/neverthrow/pull/418#discussion_r996471068) is the last thing remaining! I'm super excited to get this merged. Great job @incetarik

incetarik:
> 

@supermacro Thank you! When I'm free, I'll see this and I think it will be easy to solve and I'll be notifying you. I'm expecting this to be fixed in a day.

tests/typecheck-tests.ts
incetarik: @supermacro I have fixed this pretty easily actually, see the line `462` and `432` in `result.ts` file.

I have done this only for the case you have mentioned. If you want to apply this logic to other places as well, simply change `EmptyArrayToNever<...>` to `EmptyArrayToNever<..., 1>`. This extra `1` argument will cause to keep `[never, ...never[]]`-like array as it is. Add `1` here if you want to get precise type such as `[never, never, never]` instead of `never` which was the previous default that would be occurring if all of the elements of the array is `never`.

supermacro (approved):

supermacro:
I will be doing a review of documentation and then creating the release today / tomorrow. Thank you for your contribution @incetarik ... sending over bounty payment today :slightly_smiling_face:

PR #425, opened 2022-10-04
Let fromSafePromise properly infer err type from usage

### LINKED ISSUE #407: fromSafePromise error type must be "never"
```
const r = ResultAsync
    .fromSafePromise(new Promise<number>((resolve) => resolve(123)))
    .andThen((num) => {
        return errAsync(num);
    });
```
its type is `ResultAsync<never, unknown>` instead of `ResultAsync<never, number>`
because fromSafePromise returns `ResultAsync<type, unknown>` and "unknown" is kinda heavier than anything else I guess.
so everywhere we must write it as `fromSafePromise<type, never>` to make it work

ccjmne:
As suggested by @msheakoski, we simply default to using `never` as the err type of `fromSafePromise` when type inference fails (would be `unknown` otherwise).

Basically, you can choose to see your `never` as any specific type, but it isn't legal to do anything with your `unknown`.

I added @lmcsu and @msheakoski as co-authors.

Resolves #407

### DIFF
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -17,8 +17,8 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
     this._promise = res
   }
 
-  static fromSafePromise<T, E>(promise: PromiseLike<T>): ResultAsync<T, E>
-  static fromSafePromise<T, E>(promise: Promise<T>): ResultAsync<T, E> {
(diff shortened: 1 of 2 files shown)


msheakoski:
Thank you for taking the time to make this PR, @ccjmne! Your PR and test looks great!

supermacro (approved):

PR #426, opened 2022-10-17
Improve combine type to create a tuple of the input types

joelshepherd:
G'day, this PR improves the type of the `.combine()` functions to create a tuple of the input types.

In the following situation:
```ts
const result = Result.combine([ok(1234), ok("string")]);
```
This would previously result in this type:
`Result<(number | string)[], unknown>`

This change will improve this type to be more specific when a tuple of results is passed in:
`Result<[number, string], unknown>`

This change remains compatible if you pass an array of types in.
Ie, `Result.combine(arrayOfNumbers.map(x => ok(x))` would still create the type: `Result<number[], unknown>`.

I have added type tests for these situations.

_Side note: it seems prettier hasn't been run in the tests folder for some time, I have avoiding running it in this PR to prevent clutter._

### DIFF
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -587,7 +587,7 @@ describe('Utils', () => {
       it('Combines `testdouble` proxies from mocks generated via interfaces', async () => {
         const mock = td.object<ITestInterface>()
 
-        const result = await ResultAsync.combine([okAsync(mock)] as const)
+        const result = await ResultAsync.combine([okAsync(mock)])
 
         expect(result).toBeDefined()
         expect(result.isErr()).toBeFalsy()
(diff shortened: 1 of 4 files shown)


tests/index.test.ts
```diff
@@ -587,7 +587,7 @@ describe('Utils', () => {
       it('Combines `testdouble` proxies from mocks generated via interfaces', async () => {
         const mock = td.object<ITestInterface>()
 
-        const result = await ResultAsync.combine([okAsync(mock)] as const)
+        const result = await ResultAsync.combine([okAsync(mock)])
```
joelshepherd, line 590: The `as const` here is no longer necessary, as typescript will now infer the correct type!

This situation comes up regularly in my code, and is the motivation for this PR. Loose example:
```ts
Result.combine([
  User.find(...),
  Contact.find(...),
]).map(([user, contact]) => {
  // Use user and contact
});
```

joelshepherd:
I probably should have checked the PR list first, looks like #418 addresses the same issue.

PR #435, opened 2022-11-13
Fix types for Result.combine and ResultAsync.combine for arrays witho…

### LINKED ISSUE #434: Result.combine is broken for array types that are not tuple types
Since https://github.com/supermacro/neverthrow/issues/226 has been merged, `Result.combine` is broken for array types that are not tuple types:

```ts
import { Result } from "neverthrow";

const results: Result<string, string>[] = [];
const result = Result.combine(results); // inferred as Result<unknown, never>, but should be Result<string[], string>
```

Earlier, I had thought that this was actually related to the TypeScript option `strictFunctionTypes`, but it turns out, this is broken even with that option disabled.

Regardless, I see that neverthrow itself currently does not enable `strict` or `strictFunctionTypes` in its `tsconfig.json`, which is probably how this slipped through. I think in this day and age, enabling `strict` should be the default to ensure proper type safety, so I would recommend that neverthrow start using it, too :)

ghost91-:
…ut statically know length

The fix works be using the old implementation for the return types of `combine` and `combineWithAllErrors` in the case that the given array of `Result`s/`ResultAsync`s is not a tuple type.

While implementing this, I actually stumbled across another issue: The types for `combineWithAllErrors` are incorrect in the tuple case. At the moment, it is typed so that the error case also is a tuple of all the error types. But this is not what happens in reality, because not all of the results necessarily are errors. Here is an example:
```ts
const result1: Result<string, string> = ok('foo');
const result2: Result<number, number> = err(42);
const results = Result.combine([result1, result2]); // inferred as Result<[string, number], [string, number]>, but it should be Result<[string, number], (string | number)[]>, because we don't statically know how many of the results are errors.
expect(results).toEqual(err([42]));
```
There was actually a single test case that should have covered this, but it didn't catch it, because so far, we were only testing assignability of the result to the expectation, but not the other way round (and there are other test cases that expect the incorrect types...). I added checks for the other direction, too, so now we actually check for type equality.

The problem itself is however _not_ addressed by this PR and should be addressed separately. I just wanted to write it down so I don't forget. I’ll create an issue for it, when I find the time.

Fixes #434

### DIFF
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -11,6 +11,8 @@ import { Err, Ok, Result } from './'
 import {
   combineResultAsyncList,
   combineResultAsyncListWithAllErrors,
+  ExtractErrAsyncTypes,
+  ExtractOkAsyncTypes,
   InferAsyncErrTypes,
   InferAsyncOkTypes,
   InferErrTypes,
@@ -154,10 +156,18 @@ export const fromPromise = ResultAsync.fromPromise
 export const fromSafePromise = ResultAsync.fromSafePromise
 
 // Combines the array of async results into one result.
-export type CombineResultAsyncs<T> = TraverseAsync<UnwrapAsync<T>>
+export type CombineResultAsyncs<
+  T extends readonly ResultAsync<unknown, unknown>[]
+> = IsLiteralArray<T> extends 1
+  ? TraverseAsync<UnwrapAsync<T>>
+  : ResultAsync<ExtractOkAsyncTypes<T>, ExtractErrAsyncTypes<T>[number]>
 
 // Combines the array of async results into one result with all errors.
-export type CombineResultsWithAllErrorsArrayAsync<T> = TraverseWithAllErrorsAsync<UnwrapAsync<T>>
+export type CombineResultsWithAllErrorsArrayAsync<
+  T extends readonly ResultAsync<unknown, unknown>[]
+> = IsLiteralArray<T> extends 1
+  ? TraverseWithAllErrorsAsync<UnwrapAsync<T>>
+  : ResultAsync<ExtractOkAsyncTypes<T>, ExtractErrAsyncTypes<T>[number][]>
 
 // Unwraps the inner `Result` from a `ResultAsync` for all elements.
 type UnwrapAsync<T> = IsLiteralArray<T> extends 1
--- a/src/result.ts
+++ b/src/result.ts
@@ -3,6 +3,8 @@ import { createNeverThrowError, ErrorConfig } from './_internals/error'
 import {
   combineResultList,
   combineResultListWithAllErrors,
+  ExtractErrTypes,
+  ExtractOkTypes,
   InferErrTypes,
   InferOkTypes,
 } from './_internals/utils'
@@ -402,6 +404,8 @@ export type Combine<T, Depth extends number = 5> = Transpose<CollectResults<T>, 
   infer R,
 ]
   ? [UnknownMembersToNever<L>, UnknownMembersToNever<R>]
+  : Transpose<CollectResults<T>, [], Depth> extends []
+  ? [[], []]
   : never
 
 // Deduplicates the result, as the result type is a union of Err and Ok types.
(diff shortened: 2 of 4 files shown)


ghost91-:
In order to solve all of this properly, I think we should bite the bullet and accept that we need more modern TypeScript version. I now also have a variant that implements this properly (afaik), but it also relies on spreading:
```ts
// utils.ts

// Given a list of Results, this extracts all the different `T` types from that list
export type ExtractOkTypes<T extends readonly Result<unknown, unknown>[]> = {
  [idx in keyof T]: T[idx] extends Err<unknown, unknown>
    ? never
    : T[idx] extends Result<infer U, unknown>
    ? U
    : never
}

// Given a list of Results, get the first possible `T` type
export type FirstPossibleOk<T extends readonly Result<unknown, unknown>[]> = T extends [
  infer U,
  ...infer V
]
  ? U extends Ok<infer W, unknown>
    ? W
    : V extends readonly Result<unknown, unknown>[]
    ? FirstPossibleOk<V>
    : never
  : ExtractOkTypes<T>[number]

// Given a list of Results, this extracts all the different `E` types from that list
export type ExtractErrTypes<T extends readonly Result<unknown, unknown>[]> = {
  [idx in keyof T]: T[idx] extends Ok<unknown, unknown>
    ? never
    : T[idx] extends Result<unknown, infer E>
    ? E
    : never
}

// Given a list of Results, get the first possible `E` type
export type FirstPossibleErr<T extends readonly Result<unknown, unknown>[]> = T extends [
  infer U,
  ...infer V
]
  ? U extends Err<unknown, infer W>
    ? W
    : V extends readonly Result<unknown, unknown>[]
    ? FirstPossibleErr<V>
    : never
  : ExtractErrTypes<T>[number]

// Given a list of Results, get the first statically known `E` type
export type FirstKnownErr<T extends readonly Result<unknown, unknown>[]> = T extends [
  infer U,
  ...infer V
]
  ? [U] extends [Err<unknown, infer W>]
    ? W
    : V extends readonly Result<unknown, unknown>[]
    ? FirstKnownErr<V>
    : never
  : never

// result.ts

// Combines the array of results into one result.
export type CombineResults<
  T extends readonly Result<unknown, unknown>[]
> = FirstKnownErr<T> extends never
  ? FirstPossibleErr<T> extends never
    ? Ok<ExtractOkTypes<T>, never>
    : FirstPossibleOk<T> extends never
    ? Err<never, FirstPossibleErr<T>>
    : Result<ExtractOkTypes<T>, FirstPossibleErr<T>>
  : Err<never, FirstPossibleErr<T>>

// Combines the array of results into one result with all errors.
export type CombineResultsWithAllErrorsArray<
  T extends readonly Result<unknown, unknown>[]
> = FirstKnownErr<T> extends never
  ? FirstPossibleErr<T> extends never
    ? Ok<ExtractOkTypes<T>, never>
    : FirstPossibleOk<T> extends never
    ? Err<never, ExtractErrTypes<T>[number][]>
    : Result<ExtractOkTypes<T>, ExtractErrTypes<T>[number][]>
  : Err<never, ExtractErrTypes<T>[number][]>
```

Am I missing anything?

supermacro:
Thanks for opening a PR @ghost91- , I will be reviewing this week. I also want to cc @incetarik since he worked on a lot of related code here + he's also involved in the associated issue #434

incetarik:
> Thanks for opening a PR @ghost91- , I will be reviewing this week. I also want to cc @incetarik since he worked on a lot of related code here + he's also involved in the associated issue #434

Thanks! I'm working on something else that keeps me busy but I think I'll have time to check this out and I think we can fix this together! But really, I don't know when I can be free for this. Apart from this, I'm also interested in the another issue that was about `map` or something I don't recall it well. So I wish the best for the community.

lukebrandonfarrell:
This is a positive improvement, as I usually run into typing issues with combine in all our applications. Hoping this get merged soon 🤞

supermacro (approved):
Took quite a while to get to this - thank you for your patience and contribution!

PR #446, opened 2023-01-22
Fix a test to be in an intended way

untidy-hair:
Hi 👋 Not a big deal, but I think the tests were intended like the changes here, altho the original test was not necessarily incorrect. 
As a positive result, `expect(newResult._unsafeUnwrap()).toStrictEqual('Very Nice!')` was able to be added. (For the original setup, the equivalent would have been `expect(newResult._unsafeUnwrap()).toStrictEqual(ok('Very Nice!'))` which would look weird.)

### DIFF
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -142,7 +142,7 @@ describe('Result.Ok', () => {
       // network calls
       // disk io
       // etc ...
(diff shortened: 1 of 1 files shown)


supermacro (approved):
Thank you!

PR #454, opened 2023-03-14
fix(Readme): Fix syntax highlighting in examples

JoschuaSchneider:
Some examples in the Readme were missing the `from` and some used `const` instead of import, which messed with the syntax highlighting.


supermacro (approved):
ty 🙌

PR #456, opened 2023-03-24
feat: add a "tap"-method to perform side effects

lauhon:
Hi,

created this to run side effects like logging.

Name "tap" is inspired by  [rxjs tap](https://www.learnrxjs.io/learn-rxjs/operators/utility/do)


lauhon:
Hey, @supermacro what do you think about this?

supermacro:

src/result-async.ts
```diff
@@ -68,6 +68,19 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
...
   }
 
+  tap(f: (t: T) => void | Promise<void>): ResultAsync<T, E> {
+    return new ResultAsync(
+      this._promise.then(async (res: Result<T, E>) => {
+        if (res.isErr()) {
+          return new Err<T, E>(res.error)
+        }
+
+        await f(res.value)
```
supermacro, line 78: What if the promise fails? Shouldn't this be wrapped in a try / catch? `ResultAsync.fromPromise` could also be used here, which requires a handler function in the event the promise fails.

tests/index.test.ts
```diff
@@ -709,7 +762,7 @@ describe('ResultAsync', () => {
       expect(mapAsyncFn).toHaveBeenCalledTimes(1)
     })
 
-    it('Skips an error', async () => {
+    it('Skips an error when mapping an asynchronous value', async () => {
```
supermacro, line 765: :+1:

supermacro (changes requested):
## TODOs

- I had a question about handling failed promises for `.tap` on `ResultAsync`.
- Update README to include docs on `.tap` for both `Result` and `ResultAsync`

src/result-async.ts
lauhon: Good question, I'm honestly not sure. Maybe that's a good indicator that I should add a test that covers that(?)

I pretty much copied the `map` implementation. The only difference is that instead of returning the result of the passed function, I execute the function and then return the previous value.

If I add a try-catch to the tap function, shouldn't I also add one to the `map` and `mapErr` functions? 🤔

supermacro:
I think we're good to close this since https://github.com/supermacro/neverthrow/pull/467 was merged 

cc @m-shaka

PR #467, opened 2023-04-03
andTee() / andThrough() to handle side-effects a bit easier (from issue #445)

### LINKED ISSUE #445: Add andTee() implementation
Hi @supermacro and all the contributors, I love neverthrow, thank you for this great library!

In the ideal functional world, we don't want side-effects but in reality, we do need to handle side effects. When we do so, it is usually a good idea to separate pure functions from side effect functions cleanly. And in some of those cases, you want to use the result of the previous computation after you call a side effect function. Of course we can do so with the current API, too, but it would be sleek if we have something like `andTee()` which we would be able to use as follows:

```typescript
validate(userInfoInput).andThen(buildUser).andTee(persistUser).andThen(notifyUserRegistration)....
```
or perhaps for logging
```typescript
doSomeEventProcess(input).andTee(logEvent).andThen(doNextThing)....
```

The potential signature would be:
```typescript
andTee<F>(f: (t: T) => Result<unknown, F>): Result<T, E | F> 
// Typically unknown above is void
// In case of failure, it passes along the error F instead.
```

I think we can have variations such as `andSafeTee()` and `asyncAndTee()`, altho there might be better naming than "tee".

I found this library just about a month ago so hopefully I'm not missing anything.
I will be happy to make a PR if this idea sounds a good addition :)

untidy-hair:
closes #445 

Hi @supermacro! 
I opened [an issue with a suggestion](#445 ) about 2.5 months ago (Jan 17) but I think you have been very busy and I thought maybe a real code would save your time :)

I got a feedback comment in the original issue thread and I ended up having these two methods.

1. andTee()
```typescript
andTee(f: (t: T) => unknown): ResultAsync<T, E>
// Usage: validate() below returns ResultAsync
validate(userInfoInput).andThen(buildUser).andTee(logEvent).andThen(notifyUserRegistration)....
```
2. andThrough()
```typescript
andThrough<F>(f: (t: T) => ResultAsync<unknown, F>): ResultAsync<T, E | F>
// Usage: validate() below returns ResultAsync
validate(userInfoInput).andThen(buildUser).andThrough(persistUser).andThen(notifyUserRegistration)....
```

Both return the original input `okAsync<T>` upon success, but in case of error, `andTee()` ignores the error whereas `andThrough()` returns the new error `errAsync<F>`. The former is suitable for such as logging when you don't want to stop the main logic upon trivial side-effect failure and the latter is suitable when you want to stop the main logic upon a critical side-effect failure such as DB insertion error.

Origianlly I was going to use  the name `andTee` for no.2 above that passes error value in case of an error but considering the original meaning of "tee", even the error should not come back to the main "pipe" so I used the name for no.1 above instead. (But as I wrote in the original issue, there could be better names.)
 
And was also written in the original thread, this can be done with the current APIs but these two new methods will make it possible to write and understand each step of the logics more vividly (IMO) and I personally have cases I want to rewrite code with `andThrough()` for DB persistence as I explained here :)

Thanks in advance!


brndt:
@supermacro can you take a look on this please? It's quite useful feature to avoid boilerplate

brndt:
@supermacro bump :)

supermacro:

guilherme-sato (approved):

untidy-hair:
Hi @supermacro , have you had a chance to look at this PR? Thank you!

tests/typecheck-tests.ts
```diff
@@ -310,6 +436,155 @@ type CreateTuple<L, V = string> =
...
+        code: number
+      }
+
+      type Expectation = ResultAsync<number, MyError | string[]>
+
+      const result: Expectation = ok<number, MyError>(123)
+        .asyncAndThrough((val) => errAsync<string, string[]>(['oh nooooo']))
+    });
+
+    // (function it(_ = 'Infers error type when returning disjoint types (native scalar types)') {
```
m-shaka, line 471: This test should be executed

src/result.ts
```diff
@@ -303,6 +343,10 @@ export class Ok<T, E> implements IResult<T, E> {
     return f(this.value)
   }
 
+  asyncAndThrough<F>(f: (t: T) => ResultAsync<unknown, F>): ResultAsync<T, F> {
+    return f(this.value).map<T>((_value: unknown) => this.value)
+  }
```
m-shaka, line 348:
```suggestion
  asyncAndThrough<R extends ResultAsync<unknown, unknown>>(
    f: (t: T) => R,
  ): ResultAsync<T, E | InferAsyncErrTypes<R>>
  asyncAndThrough<F>(f: (t: T) => ResultAsync<unknown, F>): ResultAsync<T, E | F>
  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
  asyncAndThrough(f: (t: T) => ResultAsync<unknown, unknown>): any {
    return f(this.value).map(() => this.value)
  }
```
The signature should be overloaded and the error type should be union like `andThrough`.
If so, you can remove this comment out
https://github.com/supermacro/neverthrow/pull/467/files?w=1#diff-5e6df76f1ff5c2ddfbb0bee2e515236794778a429792a91d3b853bfecc884c16R471-R485

src/result.ts
```diff
@@ -375,6 +427,10 @@ export class Err<T, E> implements IResult<T, E> {
     return errAsync<U, E>(this.error)
   }
 
+  asyncAndThrough<F>(_f: (t: T) => ResultAsync<unknown, F>): ResultAsync<T, E> {
```
m-shaka, line 430:
```suggestion
  asyncAndThrough<F>(_f: (t: T) => ResultAsync<unknown, F>): ResultAsync<T, E | F> {
```

m-shaka (changes requested):
Thank you for your contribution
Can you add docs about new features to README?

m-shaka:
@supermacro Hey Gio, I think `andTee` is a nice name, but what do you think of the name `andThrough`? Do you have any other ideas?

untidy-hair:
Hi @m-shaka and @supermacro , I updated the PR based on the suggestions as well as README. Thanks

supermacro:
Both names are ok with me! Don't think it's fair for me to come in so late and opine strongly here. Great job everyone .. and thanks for your patience :)

supermacro (approved):

PR #482, opened 2023-06-14
Update .npmignore

DaviDevMod:
🤗


DaviDevMod:
Sorry, I wasn't aware that the push to my fork would have been automatically picked up by the PR.
I just wanted to show this other approach of defining what to include rather than what to exclude.
Please let me know if you prefer to keep the .npmignore and I'll revert the changes.

DaviDevMod:
By the way, here are the docs for the [files](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files) field.

You can verify that the result would be the same by running `npm pack --dry`.

PR #500, opened 2023-10-26
Pass typing-helpers to safeTry's body

tsuburin:
Edit: The descriptions for `ok` and `err` in this comment is wrong. Please see the following comments.

With `safeTry`, sometimes we need to specify type parameters to pass type-checks.
For example, the following will not pass type-checks because the body's return type (`T`) is inferred to be `{ type: string }`
```typescript
safeTry<{ type: "foo" }, unknown>(function*() {
    return ok({ type: "foo" })
})
```
A naive solution is to specify the body's type parameters like
```typescript
safeTry<{ type: "foo" }, unknown>(function*(): Generator<Err<never, unknown>, Result<{ type: "foo" }, unknown>> {
    return ok({ type: "foo" })
})
```
but this requires many keystrokes and obviously cumbersome.

With this PR, properly typed helpers will be passed to the body so that we will not be required to specify type parameters.
For example,
```typescript
safeTry<{ type: "foo" }, unknown>(function*({
  ok  // This `ok` is `ok<{ type: "foo" }, unknown>`, so using this we don't need to specify type parameters anymore
}) {
    return ok({ type: "foo" })
})
```

The helpers are `ok`, `err` and `error`. Please see the comments on the code.

This is not a breaking change.

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -126,9 +126,9 @@ type SafeTryBoundedHelpers<T, E> = {
    * which is not compatible with the enclosing safeTry's `E`: { type: "foo" }.
    * ```typescript
    * safeTry<unknown, { type: "foo" }>(function*() {
-   *   yield* err(undefeind).mapErr(() => ({
+   *   yield* err(undefined).mapErr(() => ({
    *     type: "foo"
-   *   }))
+   *   })).safeUnwrap()
    *   return ok("something")
    * })
    * ```
@@ -139,7 +139,7 @@ type SafeTryBoundedHelpers<T, E> = {
    * safeTry<unknown, { type: "foo" }>(function*({ error }) {
    *   yield* err(undefined).mapErr(() => error({
    *     type: "foo"
-   *   }))
+   *   })).safeUnwrap()
    *   return ok("something")
    * })
    * ```
--- a/src/result.ts
+++ b/src/result.ts
@@ -65,6 +65,89 @@ export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
 export const err = <T = never, E = unknown>(err: E): Err<T, E> => new Err(err)
 
 /**
+ * Passed to safeTry's body for helping type-checkers to infer body's types.
+ */
+type SafeTryBoundedHelpers<T, E> = {
+  /**
+   * The same function as neverthrow's `ok` with its `T` and `E` bound to
+   * those of the enclosing safeTry.
+   * Intended to be used to help type-checkers to infer the `Ok`'s value type.
+   *
+   * The following will not pass type-check,
+   * because the body's `T` is inferred to be `{ type: string }`,
+   * which is not compatible with the enclosing safeTry's `T`: `{ type: "foo" }`.
+   * ```typescript
+   * safeTry<{ type: "foo" }, unknown>(function*() {
+   *   // This `ok` is what is exported by neverthrow
+   *   return ok({ type: "foo" })
+   * })
+   * ```
+   *
+   * You can use this helper to safely assert that
+   * the value type is the same as the enclosing safeTry's `T`.
+   * ```typescript
+   * safeTry<number, { type: "foo" }>(function*({ ok }) {
+   *   // This `ok` is what is passed to as the body's parameter
+   *   return ok({ type: "foo" })
+   * })
+   * ```
+   */
+  ok: (value: T) => Ok<T, E>
+  /**
+   * The same function as neverthrow's `err` with its `T` and `E` bound to
+   * those of the enclosing safeTry.
+   * Intended to be used to help type-checkers to infer the `Err`'s error type.
+   *
+   * The following will not pass type-check,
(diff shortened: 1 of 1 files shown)


ghost91-:
The example you have given [_does_ pass type checks](https://www.typescriptlang.org/play?ts=5.2.2#code/JYWwDg9gTgLgBAbzhA1gGjgZwIYDMCmAKlAJ5wC+cuUEIcA5AHb4Bu+sAFjQO70DcAKAEBjCI0zwo+TAFcANvAC8WPEVIAeJDBJh8ALjgAiXBAiGKGGYxSMI3RgD4AFLivCYwMQCo4TgJSIAnDBcFIwMlCMyChOWjr6RiZmFH6C5KkCQA) for me. Maybe it depends on the version of TypeScript that's being used?

src/result.ts
```diff
@@ -64,6 +64,89 @@ export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
...
+   * The following will not pass type-check,
+   * because the body's `E` is inferred to be { type: string },
+   * which is not compatible with the enclosing safeTry's `E`: { type: "foo" }.
+   * ```typescript
+   * safeTry<unknown, { type: "foo" }>(function*() {
+   *   yield* err(undefeind).mapErr(() => ({
+   *     type: "foo"
+   *   }))
+   *   return ok("something")
+   * })
```
ghost91-, line 133:
```suggestion
   * safeTry<unknown, { type: "foo" }>(function*() {
   *   yield* err(undefined).mapErr(() => ({
   *     type: "foo"
   *   })).safeUnwrap()
   *   return ok("something")
   * })
```

src/result.ts
```diff
@@ -64,6 +64,89 @@ export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
...
+   * ```
+   *
+   * You can use this helper to safely assert that
+   * the error type is the same as the enclosing safeTry's E.
+   * ```typescript
+   * safeTry<unknown, { type: "foo" }>(function*({ error }) {
+   *   yield* err(undefined).mapErr(() => error({
+   *     type: "foo"
+   *   }))
```
ghost91-, line 142:
```suggestion
   *   })).safeUnwrap()
```

ghost91- (changes requested):

tsuburin:
@ghost91- 

> The example you have given [does pass type checks](https://www.typescriptlang.org/play?ts=5.2.2#code/JYWwDg9gTgLgBAbzhA1gGjgZwIYDMCmAKlAJ5wC+cuUEIcA5AHb4Bu+sAFjQO70DcAKAEBjCI0zwo+TAFcANvAC8WPEVIAeJDBJh8ALjgAiXBAiGKGGYxSMI3RgD4AFLivCYwMQCo4TgJSIAnDBcFIwMlCMyChOWjr6RiZmFH6C5KkCQA) for me. Maybe it depends on the version of TypeScript that's being used?

Thank you for taking your time to try reproducing it. I'm not sure I understand what happens here because I'm not used to using playground with imports, but it seems not working in those cases because it seems [this](https://www.typescriptlang.org/play?ts=5.2.2#code/JYWwDg9gTgLgBAbzhA1gGjgZwIYDMCmAKlAJ5wC+cuUEIcA5AHb4Bu+sAFjQO70DcAKAEBjCI0zwo+TAFcANvAC8WPEVIAeJDBJh8ALjgAiXBAiGMAE3wxswOQYlRgjAOYUMMxikYRujAHwAFLiewjDAYgBUcIEAlIgCcElwUjAyUIzIKIFaOvpGJmYUsYLkJQJAA) also passes the check. (I say 'seems' because I could not find any compilation outputs, and the only clue I had was that no red underlines showed up there.)

However, I also confirmed that my examples does passes with the version of Typescript of this repository by adding them in `tests/safe-try.test.ts`. Maybe it depends on version of Typescript, as you said, or perhaps I did too much while I'm simplifying examples to write this PR, but anyway I should have checked more closely, I'm sorry. One example confirmed to not pass is in #501.

I'll fix examples and documents in this PR. Thank you.

tsuburin:
@ghost91- 
I fixed documents and examples.
For `error`, I confirmed the fixed example does not pass type-checks with the version of Typescript used in this repository.
For `ok` and `err`, I could not find good examples with simple types I tried, so it seems to be my misunderstanding. Or maybe I could find by using complex types for T or E, but aside from that, I still have another good reason for those helpers, so I modified the documents. That is, if a wrong value is returned from the body, we get puzzling error messages emitted not at the place where the wrong values is returned but at the body itself. This makes engineering experience really bad, and these helpers improve it.

m-shaka:
@tsuburin Hi, thanks for your PR!
Thanks to #527 , we don't need type parameters anymore.
Do you still want your change merged? I'm not sure use cases you are thinking of, actually

tsuburin:
@m-shaka 

> Thanks to https://github.com/supermacro/neverthrow/pull/527 , we don't need type parameters anymore.

Great! I'll close this PR. Thanks

Issue #532, opened 2024-04-22
Empty Package/No JS files (neverthrow v6.2.0)


ninoseki:
Hello, first of all, thanks for creating a very useful library.

I realized that there is an issue with the latest version v6.2.0. It does not contain JS files.

```bash
$ wget https://registry.npmjs.org/neverthrow/-/neverthrow-6.2.0.tgz 
$ tar xf neverthrow-6.2.0.tgz 
$ tree package/                                                                                                                                                                  
package/
├── LICENSE
├── README.md
└── package.json

1 directory, 3 files
```

I guess there is something wrong with the build script or CI.

ninoseki:
It's strange because I can confirm dist files via `npm pack`. Maybe there was something wrong before `npm publish`?

```bash
$ git clone https://github.com/supermacro/neverthrow.git
$ cd neverthrow
$ npm pack
$ tar xf neverthrow-6.2.0.tgz                                                                                                                                               
$ tree package/                                                                                                                                                              
package/
├── LICENSE
├── README.md
├── dist
│   ├── index.cjs.js
│   ├── index.d.ts
│   └── index.es.js
└── package.json

2 directories, 6 files
```

ninoseki:
My previous comment is wrong. `npm pack` doesn't build/include dist directory. Now I think `npm publish` is done without `npm run build`.
#533 can prevent a similar issue.

supermacro:
https://github.com/supermacro/neverthrow/releases/tag/v6.2.1

PR #533, opened 2024-04-22
fix: set prepack to ensure the build before pack/publish

ninoseki:
Set `prepack` (see https://docs.npmjs.com/cli/v10/using-npm/scripts for details) to ensure the build before `pack`/`publish` to prevent an issue like #532.


supermacro (approved):
:pray:

Issue #525, opened 2023-12-29
"Finally" functionality


scarabcoder:
I don't know if there have changes made since issue #254, but I am looking for the same functionality.

My specific use case is that I'm looking to avoid race conditions where concurrent calls are made to the same function, which operates with chained `.andThen()` calls. To do that, I'm using the library async-mutex, which lets you create a lock instance, then wait/lock/release it in situations where you don't want concurrency. 

It hinges on the fact that no matter what, you are calling `release()`, so at the end of my chained andThen() calls it would be ideal to have it `release()` regardless of whether the ResultAsync ends up being a success or failure. However, I also don't want to modify the return type of the ResultAsync chain. 

Currently, my workaround is to declare my AsyncResult chain as a variable, call `.then()` on that AsyncResult chain and have it call `release()` in both the failure and success callbacks, then return the original AsyncResult. That looks like this:

```TypeScript
const result = ResultAsync.fromSafePromise(
      lock.acquire(),
    ).andThen(() => <additional logic>);
    
    // Ensure that the lock is released even if the transaction fails
    result.then(
      () => {
        lock.release();
      },
      () => {
        lock.release();
      },
    );

    return result;
```

This is a little annoying, the preferred syntax would be something like:

```TypeScript
return ResultAsync.fromSafePromise(
      lock.acquire(),
    ).andThen(() => <additional logic>)
    .finally(() => lock.release());
```

Where `finally` wouldn't modify the error or success value types being returned.

scarabcoder:
This solution is specific to my use case with async-mutex, but here's a util function I came up with to make the syntax a bit more developer-friendly (while also preserving types):

```TypeScript
export const withLock = <T, E>(
  lock: Mutex,
  fn: () => ResultAsync<T, E>,
): ResultAsync<T, E> => {
  const result = ResultAsync.fromSafePromise(lock.acquire()).andThen(() => {
    return fn();
  });

  result.then(
    () => {
      lock.release();
    },
    () => {
      lock.release();
    },
  );
  return result;
};
```

Usage:
```TypeScript
const result = await withLock(lock, () => doMyDbOperation()); // waits for a lock, locks, then runs doMyDbOperation(), before unlocking and returning the AsyncResult of doMyDBOperation
result.isOk();
result.value // the original response of doMyDbOperation
```

scarabcoder:
[...]

PR #536, opened 2024-05-12
feat(andFinally): add andFinallyFunctionality

kishore03109:
## Description

Add 'finally' functionality to neverthrow. Overall leads to clearer typing with less verbose code. The mental model is similar to the native [finally](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally). 

This PR attempts to solve 2 main issues. Consider the example below: 

```
    const doesResourceExist = async (): Promise<true> => {
      const simulateError = () => Math.random() < 0.33
      if (simulateError()) {
        throw new Error('Unrecoverable error')
      }
      if (simulateError()) {
        throw new Error('404 Not found error')
      }
      return true
    }

    const safeFunction: ResultAsync<true, ResultAsync<boolean, unknown>> = ResultAsync.fromPromise(
      doesResourceExist(),
      (error) => error,
    ).mapErr((error) => {
      if (`${error}` === '404 Not found error') {
        return okAsync(false)
      }
      return errAsync(error)
    })
```
Notice the return type of the `safeFunction` is typed to an unintuitive type. Not sure if there was a convenient way for error recovery while mutating the `Ok` type. With `andFinally`, there is easier control over the `Ok` type during error recovery as such:

```
    const improvedSafeFunction: ResultAsync<boolean, unknown> = ResultAsync.fromPromise(
      doesResourceExist(),
      (error) => error,
    ).andFinally((error) => {
      if (`${error}` === 'Not found error') {
        return okAsync(false)
      }
      return errAsync(error)
    })
```

Additionally, this solves  #525. As documented in the issue, the below code becomes less verbose. 

### Before
```
    const doStuff: Result<string, never> = ok('start db connection')
      .andThen(() => {
        // do stuff that might return an err()
        const result: Result<string, string> = ok('do something')
        return result
      })
      .andThen(() => {
        return ok('close db connection')
      })
      .orElse(() => {
        return ok('close db connection')
      })

```

### After 
```
    const doStuff: Result<string, never> = ok("start db connection").andThen(() => {
      // do stuff that might return an err()
      const result: Result<string, string> = ok("do something")
      return result
    }).andFinally(() => {
      return ok("close db connection")
    })
```

## Testing

Have added some minimal tests to capture correct behaviour.

### DIFF
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -131,6 +131,27 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
     )
   }
 
+  andFinally<R extends Result<unknown, unknown>>(
+    f: (t: T) => R,
+  ): ResultAsync<InferOkTypes<R>, InferErrTypes<R>>
+  andFinally<R extends ResultAsync<unknown, unknown>>(
+    f: (t: T) => R,
+  ): ResultAsync<InferAsyncOkTypes<R>, InferAsyncErrTypes<R>>
+  andFinally<U, F>(f: (t: T) => Result<U, F> | ResultAsync<U, F>): ResultAsync<U, F>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+  andFinally(f: any): any {
+    return new ResultAsync(
+      this._promise.then((res) => {
+        if (res.isErr()) {
+          return f(res.error)
+        }
+
+        const newValue = f(res.value)
+        return newValue instanceof ResultAsync ? newValue._promise : newValue
+      }),
+    )
+  }
+
   orElse<R extends Result<T, unknown>>(f: (e: E) => R): ResultAsync<T, InferErrTypes<R>>
   orElse<R extends ResultAsync<T, unknown>>(f: (e: E) => R): ResultAsync<T, InferAsyncErrTypes<R>>
   orElse<A>(f: (e: E) => Result<T, A> | ResultAsync<T, A>): ResultAsync<T, A>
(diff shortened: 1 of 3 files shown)


braxtonhall (commented):
Nice! I think it might be even nicer though if the original values are not lost when calling `.andFinally`

As a user, I think I expect that `finally` used for side effects, and probably shouldn't change the type of the result.

For example with promises (linked in your PR description):
```typescript
(async () => {
	const resolveFinally = await Promise.resolve("foo")
		.then((a) => a)
		.finally(() => "bar");

	const rejectFinally = await Promise.reject("foo")
		.catch((a) => a)
		.finally(() => "bar");

	console.log(resolveFinally, rejectFinally); // prints '"foo" "foo"'
})();
```
I guess the `try`/`finally` in JavaScript is a counter-example, because the `finally` block can overwrite the `return` in the `try` or `catch` block, but in the case of no `return` statement in the `finally` block, the original `return` value in the `try` or `catch` block is used.

It would be nice to be able to still observe that an error occurred even after the `andFinally` callback executes. Here's an example that looks just like JavaScript promises:
```typescript
const doStuff: Result<string, number> = ok('start db connection')
	.andThen(() => {
		// do stuff that might return an err()
		const result: Result<string, string> = ok('do something')
		return result
	})
	.andFinally(() => {
		// close the db connection
	})
	.orElse(() => {
		// trigger some alarm!
		return err(1);
	})
```
... although this just looks suspiciously like #456

kishore03109:
Closing because some of the comments made by @macksal & @braxtonhall were valid, and was not construct a neat way to address their concerns

macksal:
Hi @kishore03109 I've added a new implementation in #588 with a slightly different design, please consider reviewing if you find it useful!

PR #552, opened 2024-07-04
WIP  Proposal: adding [Symbol.dispose] to check unhandled Result

pnodet:
Typescript 5.2 introduced the `using` keyword.
One thing that I find hard with neverthrow is checking for unhandled result.
The eslint plugin brings helpful linter errors but I wanted to try implementing [Symbol.dispose] and see if it can help the DX.


pnodet:
Maybe this can also be helpful for issues like #525

supermacro:
I think having improvements to signal that you have unhandled / swalled errors in your stack would be awesome.

Are you able to provide examples of how this API would address this problem?

pnodet:
@supermacro Sorry, rebased this branch with upstream commits and GitHub closed it automatically.

This idea behind this is to leverage the Symbol.dispose new functionality.

```ts
const someFunc = () => {
  using anExampleResult = aFunctionReturningAResult();
}
```

for example this would throw an `UnhandledResultError` because at the end of `someFunc` the `anExampleResult` result has not been unwrap or match or another neverthrow result's method

pnodet:
tbh this was more a wild idea that a real proposal bc I see several issues (people forgetting the `using` keyword, having error thrown in production…)

I'd be happy to improve this or keep working on it if you or others think it would be a great addition tho!

supermacro:
@m-shaka @dmmulroy what are your thoughts?

pnodet:
Also curious to know @mattpocock if you have thoughts on how to improve this or examples of more complex Symbol.dispose use cases

mattpocock:
Don't see how `using` differs from just `const` here - `using` is meant for tidying up a resource when it leaves scope.

pnodet:
With `using` we can leverage Symbol.dispose to do something like :

https://github.com/supermacro/neverthrow/pull/552/files#diff-fe54c0b0441ef7719a22666a2bb171f2db9e99c0ed52dd8fec96a9882e7c842dR405-R409

Screenshot 2024-09-06 at 22 10 26

mattpocock:
@pnodet Gotcha, that makes sense. Feels possible, but I'm not sure how intuitive the API will feel to most users.

m-shaka:
I don't know if it's a use case of `using`.
`using` can be used to manage "resource" with a specific lifetime. It sounds strongly connected to side-effects, which `neverthrow` has nothing to do with
> Resource — An object with a specific lifetime, at the end of which either a lifetime-sensitive operation should be performed or a non-garbage-collected reference (such as a file handle, socket, etc.) should be closed or freed.

https://github.com/tc39/proposal-explicit-resource-management

mattpocock:
Yeah, this feels like `.unsafeUnwrap` with extra steps.

supermacro:
Identifying if a result is used or not should be done statically (prob through a linter) rather than at runtime, no? Closing for now

Issue #561, opened 2024-08-16
safeTry with async generator should return ResultAsync


mattpocock:
```ts
import { safeTry, ok } from "neverthrow";

safeTry(async function* () {
  return ok("Hey");
}).map((str) => str.toUpperCase());
```

This should be possible, but isn't.

This makes the lib less composable.

supermacro:
Fixed by #562

mattpocock:
Siiiiiick I'm really excited about this.

PR #562, opened 2024-08-16
Return ResultAsync<T, E> from safeTry instead of Promise<Result<T, E>> for better composability

sharno:


.changeset/safeTryAsyncResult.md
```diff
@@ -0,0 +1,5 @@
+---
+'neverthrow': patch
```
mattpocock, line 2: Would likely need to be a minor?

src/result.ts
```diff
@@ -104,26 +104,24 @@ export function safeTry<
...
-): Promise<
-  Result<
-    InferOkTypes<GeneratorReturnResult>,
-    InferErrTypes<YieldErr> | InferErrTypes<GeneratorReturnResult>
-  >
+): ResultAsync<
+  InferOkTypes<GeneratorReturnResult>,
+  InferErrTypes<YieldErr> | InferErrTypes<GeneratorReturnResult>
 >
 export function safeTry<T, E>(
```
mattpocock, line 117: Could this become a set of function overloads?

I.e. one branch takes a generator, returns a Result. The other takes an async generator, returns ResultAsync.

mattpocock:
Appears to be a fix for #561

sharno:
@mattpocock 
It missed this comment written right above the code that this PR is changing
```
// NOTE:
// Since body is potentially throwable because `await` can be used in it,
// Promise<Result<T, E>>, not ResultAsync<T, E>, is used as the return type.
```

not sure if I understand it fully

src/result.ts
sharno: It's already like this

```ts
export function safeTry<T, E>(body: () => Generator<Err<never, E>, Result<T, E>>): Result<T, E>
```
and
```ts
export function safeTry<T, E>(
  body: () => AsyncGenerator<Err<never, E>, Result<T, E>>,
): ResultAsync<T, E>
```

mattpocock:
@sharno Agree. Neverthrow doesn't guarantee that a function will never throw. So this feels overly defensive on the library's part.

supermacro (approved):

supermacro:
@sharno looks good, but note that types need to be updated on some tests

sharno:
updated and fixed the tests 👍

supermacro (approved):

PR #563, opened 2024-08-18
Made err() infer strings narrowly for easier error tagging

mattpocock:
I would like `err` to infer strings more narrowly. This would make it easier to create unions out of possible string errors.

```ts
const example = (num: number) => {
  if (num < 0) return err('Too low');
  if (num > 100) return err('Too high');

  return ok(num);
}
```

This would infer as `Ok<number> | Err<'Too low'> | Err<'Too high'>`.

Let me know if this is desirable.

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -62,7 +62,11 @@ export type Result<T, E> = Ok<T, E> | Err<T, E>
 
 export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
 
-export const err = <T = never, E = unknown>(err: E): Err<T, E> => new Err(err)
+export function err<T = never, E extends string = string>(err: E): Err<T, E>
+export function err<T = never, E = unknown>(err: E): Err<T, E>
+export function err<T = never, E = unknown>(err: E): Err<T, E> {
+  return new Err(err)
+}
 
 /**
  * Evaluates the given generator to a Result returned or an Err yielded from it,
--- a/tests/typecheck-tests.ts
+++ b/tests/typecheck-tests.ts
@@ -338,11 +338,11 @@ type CreateTuple<L, V = string> =
     });
 
     (function it(_ = 'combines only err results into one') {
-      type Expectation = Result<[ never, never ], number | string>;
+      type Expectation = Result<[ never, never ], number | 'abc'>;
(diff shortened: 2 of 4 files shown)


src/result.ts
```diff
@@ -62,7 +62,11 @@ export type Result<T, E> = Ok<T, E> | Err<T, E>
 
 export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
 
-export const err = <T = never, E = unknown>(err: E): Err<T, E> => new Err(err)
+export function err<T = never, E extends string = string>(err: E): Err<T, E>
```
mattpocock, line 65: Can do this with an overloaded type if you'd prefer, but figured this was the easiest way to make this change.

tests/typecheck-tests.ts
```diff
@@ -999,6 +999,24 @@ type CreateTuple<L, V = string> =
       });
     });
   });
+
+  (function describe(_ = 'err') {
```
mattpocock, line 1003: Wasn't sure where this test should go, so made its own block.

tests/typecheck-tests.ts
```diff
@@ -999,6 +999,24 @@ type CreateTuple<L, V = string> =
...
+    });
+
+    (function it(_ = 'infers the error type widely when it is not a string') {
+      type Expectation = Result<never, { abc: number }>
+
+      const result = err({ abc: 123 })
+
+      const assignableToCheck: Expectation = result;
+    });
+  })
```
mattpocock, line 1019: Side note - prettier can't run on this file because it's an outdated version and there's an 'import type'.

When I did try to run prettier on it, a LOT of changes were made, probably to do with the semicolons.

supermacro: @m-shaka mentioned we may want to use `biome` instead of prettier, but perhaps if we just update prettier this issue goes away?

supermacro (approved):

supermacro:
Thanks for the contribution @mattpocock :)

macksal:
Just found this in the blame while working on #584. Nice work @mattpocock. It might be useful for `ok/okAsync/errAsync` to have the same overload.

PR #565, opened 2024-08-19
Added failing type test for ResultAsync.fromThrowable

mattpocock:


m-shaka:
It can't be fixed, I suppose

> When inferring from a type with multiple call signatures (such as the type of an overloaded function), inferences are made from the last signature (which, presumably, is the most permissive catch-all case). It is not possible to perform overload resolution based on a list of argument types.
https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#conditional-type-constraints

See also https://github.com/microsoft/TypeScript/issues/43187

PR #567, opened 2024-08-20
Moved repo to strict: true and restructured tsconfig.jsons to use project references

mattpocock:
No changeset needed, not a user-facing change.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,13 +1,14 @@
 {
   "compilerOptions": {
-    "target": "es2022",
+    "target": "ES2015",
     "module": "ESNext",
     "moduleResolution": "Node",
     "strict": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "lib": [
-      "es2022"
+      "ES2016",
+      "ES2017.Object"
     ],
     "skipLibCheck": true,
     "esModuleInterop": true,


package.json
```diff
@@ -9,11 +9,10 @@
     "dist"
   ],
   "scripts": {
-    "test": "jest && npm run test-types",
-    "test-types": "tsc --noEmit -p ./tests/tsconfig.tests.json",
+    "test": "jest",
     "lint": "eslint ./src --ext .ts",
     "format": "prettier --write 'src/**/*.ts?(x)' && npm run lint -- --fix",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc -b",
```
mattpocock, line 15: Now, this typecheck command checks the entire repo, including the typecheck tests.

tests/index.test.ts
```diff
@@ -969,9 +962,9 @@ describe('ResultAsync', () => {
     it('Accepts an error handler as a second argument', async () => {
       const example = ResultAsync.fromThrowable(
         () => Promise.reject('No!'),
-        (e) => new Error('Oops: ' + e)
+        (e) => new Error('Oops: ' + e),
```
mattpocock, line 965: Looks like prettier changes in this file.

tsconfig.json
```diff
@@ -1,31 +1,25 @@
...
-    "src/**/*.ts"
+    "src/**/*.ts",
   ],
-  "exclude": [
-      "node_modules",
-      "**/*.spec.ts"
+  "exclude": [],
+  "references": [
+    {
+      "path": "./tsconfig.tests.json"
```
mattpocock, line 22: This is now referenced in the root tsconfig, meaning both can be run simultaneously.

tsconfig.json
```diff
@@ -1,31 +1,26 @@
...
-      "noImplicitAny": true,
-      "sourceMap": false,
-      "noUnusedLocals": true,
-      "noUnusedParameters": true,
-      "strictNullChecks": true,
-      "strictFunctionTypes": true,
-      "declaration": true,
-      "baseUrl": "./src",
-      "lib": [
-        "dom",
```
mattpocock, line 16: Removed the DOM types, not needed here.

tsconfig.json
```diff
@@ -1,31 +1,26 @@
...
   "compilerOptions": {
-      "target": "es2015",
-      "module": "ESNext",
-      "moduleResolution": "Node",
-      "strict": false,
-      "noImplicitAny": true,
-      "sourceMap": false,
-      "noUnusedLocals": true,
-      "noUnusedParameters": true,
-      "strictNullChecks": true,
```
mattpocock, line 11: Slimmed down the tsconfig by flags inferred by other flags.

.vscode/settings.json
supermacro, line 1: I don't think neverthrow should include editor specific files. Please remove.

src/result.ts
```diff
@@ -23,14 +23,16 @@ export namespace Result {
...
+    return ((...args: any[]) => {
       try {
         const result = fn(...args)
         return ok(result)
       } catch (e) {
         return err(errorFn ? errorFn(e) : e)
       }
-    }
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    }) as any
```
supermacro, line 35: Can you explain the rationale for these `as any` changes made here and in other files?

package.json
supermacro: Doesn't this create artifacts, though? Shouldn't it be `tsc -b --noEmit`? 

The whole intention of the `typecheck` script is to typecheck without generating artifacts.

tsconfig.tests.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
+    "declaration": true,
+    "composite": true,
+    "outDir": "./tests/dist",
+    "noEmit": false
```
supermacro, line 9: Shouldn't this be `"noEmit": true`?

supermacro (changes requested):

package.json
mattpocock: No, noEmit has been moved to the config file - this is so that users running `tsc` without using the command don't accidentally generate artifacts.

src/result.ts
mattpocock: Moving to strict: true revealed some errors that I didn't have time to debug. I suggest we 'as any' them now, and fix them in a followup PR.

tsconfig.tests.json
mattpocock: Using project references means that you do need to generate declaration files. This acts as a kind of cache for TS, making things run a bit faster.

These files are outputted to `./tests/dist`, as specified in `outDir`

src/result.ts
boyeln: The reason this errors and you have to `as any` it is because with this call signature you can in theory do something like:

```ts
const safeParse = fromThrowable<typeof JSON.parse, number>(JSON.parse);
      // ^ (text: string, reviver?: (this: any, key: string, value: any) => any) => Result<any, number>
```

So by allowing providing a specific generic type as E without providing an errorFn, you can get in trouble. You can then end up in a case where you try to pass `e` to the `err` in the catch block (in the second/false ternary case), but since the user might have specified a specific E (like number in the example above), you can't just give it the unknown (or any here I suppose)`e`.

The reason why `strict: true` triggers this is because previously the `errorFn` didn't get the `| undefined` from the optional notation on the parameter `errorFn?`, since those does not affect the type without `strict` turned on. So basically typescript never considered that `errorFn` could be falsy, thus never hitting the `err(e)` code path of the ternary in the catch block, until strict was turned on.

The easiest fix might be to add some overloads  to prevent using the function as I did above:

```ts
export function fromThrowable<Fn extends (...args: readonly any[]) => any>(
  fn: Fn
): (...args: Parameters<Fn>) => Result<ReturnType<Fn>, unknown>;
export function fromThrowable<Fn extends (...args: readonly any[]) => any, E>(
    fn: Fn,
    errorFn: (e: unknown) => E,
  ): (...args: Parameters<Fn>) => Result<ReturnType<Fn>, E>;
  export function fromThrowable<Fn extends (...args: readonly any[]) => any, E>(
    fn: Fn,
    errorFn?: (e: unknown) => E,
  ): (...args: Parameters<Fn>) => Result<ReturnType<Fn>, E | unknown> {
    // implementation
}
```

I don't know if it really makes any practical difference here, but I believe it would be more correct to do `catch (e: unknown)` to have `e` typed as unknown and not the default `any`.

This could of course happen in a different PR. And I suppose in theory changing the call signature as I proposed might be considered a breaking change. However, if someone is using the function as I did above they should stop doing as it probably gives you types that don't match the runtime types. So you could consider it a bug maybe 🤷

boyeln: I created a separate PR for fixing the issue: https://github.com/supermacro/neverthrow/pull/583 However, feel free to close that PR and include the changes here instead if that's better.

PR #569, opened 2024-08-20
Added arethetypeswrong to check built outputs

mattpocock:


package.json
```diff
@@ -15,8 +15,8 @@
     "format": "prettier --write 'src/**/*.ts?(x)' && npm run lint -- --fix",
     "typecheck": "tsc --noEmit",
     "clean": "rm -rf ./dist ./tmp",
-    "build": "npm run clean && rollup --config && mv tmp/*.js dist",
-    "prepack": "npm run build",
+    "build": "npm run clean && rollup --config && mv tmp/*.js dist && attw --pack .",
+    "prepublish": "npm run build",
```
mattpocock, line 19: Moved to prepublish, since attw --pack runs npm pack, creating an infinite loop with prepack

package.json
```diff
@@ -15,8 +15,8 @@
     "format": "prettier --write 'src/**/*.ts?(x)' && npm run lint -- --fix",
     "typecheck": "tsc --noEmit",
     "clean": "rm -rf ./dist ./tmp",
-    "build": "npm run clean && rollup --config && mv tmp/*.js dist",
-    "prepack": "npm run build",
+    "build": "npm run clean && rollup --config && mv tmp/*.js dist && attw --pack .",
+    "prepublishOnly": "npm run build",
```
mattpocock, line 19: Moved to prepublishOnly, since this only runs on publish instead of on pack.

attw --pack calls npm pack, so you end up in a nice infinite loop.

dmmulroy (changes requested):
Hey @mattpocock - could you explain what `attw` is and why it's a useful tool for library authors as context? 🙏

supermacro:
This package / pr is still not clicking with respect to the bigger picture PR of https://github.com/supermacro/neverthrow/pull/570

In my head I am asking myself, "why do I need `@arethetypeswrong/cli` if we already have `tsc` doing typechecking that short-circuits on a typeerror".

> It's extremely useful for checking if the exported types are correct in the package.json. I wanted to add it to make sure that my tsup changes weren't breaking anything.

So this makes me think that this PR is purely here to ascertain that the change from rollup --> tsup is working as expected and nothing else, correct?

mattpocock:
@supermacro 

`tsc` and `attw` do different things. `tsc` checks that your code type checks. `attw` checks that other packages consuming your code can consume it. It prevents errors like mis-configuring `exports`.

More info in [this article](https://www.totaltypescript.com/how-to-create-an-npm-package#5-exports-main-and-arethetypeswrongcli).

You're right, it is partially a bridging PR to ensure that `tsup` makes no meaningful changes.

But it does have its own utility. In the future, you may want to consider adding multiple entrypoints to `neverthrow` to prevent the downsides of barrel files. `attw` would ensure that `exports` would still be configured correctly.

Overall, I like as much of the user experience as possible to be testable via CI. Checking that users can consume the code feels important. But if you don't want it checked on every CI run, I can understand that.

dmmulroy:
I think that this is positive change - @supermacro and @m-shaka could we adopt this and merge it to unblock #570?

supermacro:
Replaced by https://github.com/supermacro/neverthrow/pull/618

PR #570, opened 2024-08-20
Matt/migrated-build-system-to-tsup

mattpocock:


dmmulroy (changes requested):
@mattpocock could you please add some context to why `tsup` is a good/better solution beyond just being "modern"? Thanks!

@supermacro @m-shaka

mattpocock:
The main appeal of `tsup` is that I trust the declaration files it generates. It also generates one per output type by default (which is recommended by Andrew Branch of the TS team) which means the `exports` map can be simpler.

If you're happy with your current rollup-based setup, that's fine - but it seemed a bit custom and funky so I thought a low-config, well-maintained option would make sense.

supermacro:
refraining from diving into this PR until convo for https://github.com/supermacro/neverthrow/pull/569 is clarified.

jalavosus:
@supermacro It appears that #569/#618 were merged. Any chance you could circle back to this?

PR #583, opened 2024-09-12
prevent passing a specific error type for `.fromThrowable` without providing an error mapping function

boyeln:
As discussed in https://github.com/supermacro/neverthrow/pull/567#discussion_r1754313850, the current type signature of `fromThrowable` allows users to pass a specific error value type, without providing an error mapping function that produces the provided error type:

```ts
const safeParse = Result.fromThrowable<typeof JSON.parse, number>(JSON.parse);
      // ^ (text: string, reviver?: (this: any, key: string, value: any) => any) => Result<any, number>
```

This PR tries to solve that by adding some overloads to the function, only allowing users to specify the error type when a matching error mapping function is provided.

### DIFF
--- a/.changeset/grumpy-hats-unite.md
+++ b/.changeset/grumpy-hats-unite.md
@@ -1,0 +1,5 @@
+---
+'neverthrow': minor
+---
+
+prevent passing a specific error type for `.fromThrowable` without providing an error mapping function
--- a/src/result.ts
+++ b/src/result.ts
@@ -17,18 +17,33 @@ export namespace Result {
    * arguments but returning `Ok` if successful, `Err` if the function throws
    *
    * @param fn function to wrap with ok on success or err on failure
-   * @param errorFn when an error is thrown, this will wrap the error result if provided
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  export function fromThrowable<Fn extends (...args: readonly any[]) => any>(
+    fn: Fn,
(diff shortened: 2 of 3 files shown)


.changeset/grumpy-hats-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'neverthrow': minor
```
boyeln, line 2: This might be considered a breaking change, since users that use the function like described in the PR will get red squiggles. However, I would argue that this is a "type bug", and shouldn't be considered a valid use case

src/result.ts
```diff
@@ -17,18 +17,33 @@ export namespace Result {
...
+  export function fromThrowable<Fn extends (...args: readonly any[]) => any>(
+    fn: Fn,
+  ): (...args: Parameters<Fn>) => Result<ReturnType<Fn>, unknown>
+  /**
+   * Wraps a function with a try catch, creating a new function with the same
+   * arguments but returning `Ok` if successful, `Err` if the function throws
+   *
+   * @param fn function to wrap with ok on success or err on failure
+   * @param errorFn when an error is thrown, this will wrap the error result
+   */
```
boyeln, line 31: It's not the most elegant to have to maintain two versions of the documentation, however I'm not aware of any other way of doing it. Please let me know if there's a better way!

m-shaka: You can't avoid that, I guess.
Here is the official doc comment of `Array.prorotype.concat`. The same content is written in two places.
image

m-shaka (approved):
Looks great! Thank you
Sorry for the late review

PR #584, opened 2024-09-19
Allow ok/err/okAsync/errAsync to accept zero arguments when returning void

macksal:
When implementing a function returning a result, it's often that I need to construct an `Ok<void>` object. However, it's not so ergonomic to do this. `ok` must be passed an argument, even if you explicitly specify the `void` type. This leads to passing a "dummy" value:

```ts
return ok<void>(undefined);
```

Meanwhile, in Promise-land, it's perfectly acceptable to call `Promise.resolve()` with no arguments, which returns a `Promise<void>`.

Interestingly, TS allows a parameter to be omitted if it has type `void`. However, this does not work if the type is generic and has been expanded to `void` or a union containing `void`. See https://github.com/Microsoft/TypeScript/issues/29131

This PR adds additional overloads for `ok`, `err`, `okAsync` and `errAsync` that separate out the case where the argument type is `void`. This allows the following to work as expected:

```ts
return ok<void>();
return ok();
```

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -61,10 +61,15 @@ export namespace Result {
 
 export type Result<T, E> = Ok<T, E> | Err<T, E>
 
-export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
+export function ok<T, E = never>(value: T): Ok<T, E>
+export function ok<T extends void = void, E = never>(value: void): Ok<void, E>
+export function ok<T, E = never>(value: T): Ok<T, E> {
+  return new Ok(value)
+}
 
 export function err<T = never, E extends string = string>(err: E): Err<T, E>
 export function err<T = never, E = unknown>(err: E): Err<T, E>
+export function err<T = never, E extends void = void>(err: void): Err<T, void>
 export function err<T = never, E = unknown>(err: E): Err<T, E> {
   return new Err(err)
 }
--- a/.changeset/thirty-avocados-explain.md
+++ b/.changeset/thirty-avocados-explain.md
@@ -1,0 +1,5 @@
+---
+"neverthrow": minor
+---
+
+Allow ok/err/okAsync/errAsync to accept zero arguments when returning void
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -207,11 +207,17 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
   }
 }
 
-export const okAsync = <T, E = never>(value: T): ResultAsync<T, E> =>
-  new ResultAsync(Promise.resolve(new Ok<T, E>(value)))
+export function okAsync<T, E = never>(value: T): ResultAsync<T, E>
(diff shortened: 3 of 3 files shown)


petermakeswebsites:
Would really appreciate this being implemented!

macksal:
> Would really appreciate this being implemented!

Please help by providing a code review and hopefully we can get attention from the maintainers @supermacro @m-shaka. I'm using this in production and have considered forking, but this is already a fairly niche library and I hope we could maintain some kind of central community.

petermakeswebsites (approved):

src/result.ts
```diff
@@ -61,10 +61,15 @@ export namespace Result {
...
-export const ok = <T, E = never>(value: T): Ok<T, E> => new Ok(value)
+export function ok<T, E = never>(value: T): Ok<T, E>
+export function ok<T extends void = void, E = never>(value: void): Ok<void, E>
+export function ok<T, E = never>(value: T): Ok<T, E> {
+  return new Ok(value)
+}
 
 export function err<T = never, E extends string = string>(err: E): Err<T, E>
 export function err<T = never, E = unknown>(err: E): Err<T, E>
+export function err<T = never, E extends void = void>(err: void): Err<T, void>
```
Papooch, line 72: Is there really any valid use-case for void errors?

macksal: I've never used one personally. Perhaps someone is using it as a generic "failure" while making a quick and dirty prototype.

I included the `err` overload for completeness, not for any specific use case.

I think it would be more surprising for the consumer if ok/err didn't work the same way. Do you see any downside to this?

Papooch: In my humble opinion, this is a case of YAGNI.
It feels _wrong_ to be able to return an error which is void.

petermakeswebsites: My 2c... 

I don't disagree with you @Papooch . I don't really see a use case for it either. Doesn't mean there isn't one.

However, I will say I think there's an overYAGNIation in libraries in general. So many times I've come across a feature I've wanted and it hasn't been implemented because of YAGNI. No human being can think of all the possible use cases, and in this case, it's not like adding that overload has any drawbacks.

At the end of the day though, I don't really mind either way - I'd just like to see the void Ok implemented! So whatever makes everyone else happy to expedite that process is fine by me! :)

m-shaka: That's ok for me. 
It looks like that Rust's Err accepts Unit type

m-shaka (approved):
Thank you @macksal !
@supermacro will merge the PR if it's ok for him

supermacro (approved):

macksal:
Thank you @supermacro @m-shaka for merging.

Would you consider reviewing another PR of mine, #588? It is adding a commonly requested feature including test coverage. I will appreciate any comment on whether this feature could be accepted.

pfe-nazaries:
How could this be achieved via .map method?

PR #589, opened 2024-10-09
safeTry should not require .safeUnwrap()

### LINKED ISSUE #580: [Feature request] `safeTry` should not require `.safeUnwrap()`
One of neverthrow's weakest API's is `safeTry`, combined with `.safeUnwrap`

```ts
const result = safeTry(function*() {
  const foo = yield* ok("foo").safeUnwrap();
});
```

I propose, instead, that `.safeUnwrap()` should be called whenever a `Result` is `yield*`-ed.

```ts
const result = safeTry(function*() {
  const foo = yield* ok("foo");
});
```

This matches the similar API found in [Effect.gen](https://effect.website/docs/guides/essentials/using-generators#understanding-effectgen).

To make this work, `Result` and `ResultAsync` should expose a `[Symbol.iterator]` which calls `.safeUnwrap`. Since it's rare that anyone would want to iterate over a `Result`, this API feels about as safe as `.safeUnwrap()`.

dmmulroy:
Resolves #580 

CC: @mattpocock @m-shaka @supermacro


tests/index.test.ts
m-shaka, line 1: Can you revert all the unrelated changes and commit only the relevant ones?

tests/safe-try.test.ts
m-shaka, line 1: same as above

tests/index.test.ts
dmmulroy: lmao missed that formatting absolutely nuked things. fixing this now

package.json
```diff
@@ -36,21 +37,22 @@
     "@babel/preset-typescript": "7.24.7",
     "@changesets/changelog-github": "^0.5.0",
     "@changesets/cli": "^2.27.7",
-    "@types/jest": "27.4.1",
+    "@jest/test-utils": "^0.0.0",
```
m-shaka, line 40: Why was `@jest/test-utils` added?

dmmulroy: Turns I didnt need it and have removed it. Was trying to figure out how to get Jest to support comparing objects with `*[Symbol.iterator]` with `.toBeEqual`. It ended up just needing a version bump of jest

dmmulroy:
> LGTM, but CI failed again 😢

@m-shaka I missed a `,` in the `eslint-ignore` 😭 - fixed and pushed though haha

m-shaka (approved):

supermacro (approved):

Issue #587, opened 2024-09-27
[Feature request] have `.unwrapOr()` equivalent that runs a function 


janglad:
Title, probably best to implement it as a new method as this could be breaking for people that are currently returning functions from `.unwrapOr()`. 

Example usecase specific usecase for me 

```ts

declare const getSignedInUser: () => ResultAsync<User, AuthError> 

import { redirect } from "next/navigation";
const redirectToAuth = () => {
 // Throws under the hood so has returntype of never
    redirect("/login")
}

// On page
const user = await getSignedInUser().unwrapOr(redirectToAuth)
```

but a more common use case would be having a dynamic value  as a fallback (`result.error` could be provided as argument) or one that you want to compute lazily.

If this is something other people want please leave a suggestion for naming and I can probably implement this.

paduc:
Hello @janglad 

I believe `match` can handle your use case.

```ts
const user = await getSignedInUser().match( (user) => user, (e) => { redirectToAuth() }); 
```

mattpocock:
@janglad `mapErr` does this, I think

janglad:
@mattpocock not quite, still leaves you with a Result 

```ts
interface User {
  id: string;
}

interface AuthError {
  code: string;
}

declare const getUser: () => ResultAsync<User, AuthError>;
declare const redirect: () => never;

const user = await getUser().mapErr(redirect);
//  Result<User,never>

const user2 = await getUser().match((user) => user, redirect);
//    User

```

while not super on topic for this issue this does make me wonder if a `Result<T, never>` should be typed as an `Ok<T, never>`? The `Ok` would have `.value` available right away without checking for `isOk()`, and the `never` already indicates that there can't be any error case. Altho no idea if this is actually implemented differently and might cause issues. 

@paduc that's true, but while obv a quick `val => val` isn't the biggest deal it's also not the most elegant.

paduc:
@janglad I believe what you are trying to do may be outside the scope of this lib, which is handling error types. You are trying to add behavior / actions.

There are plenty of ways to achieve this without adding to the API.

janglad:
@paduc yea agree that redirecting is def out of scope, but not sure if using a function to get the fallback is? Similar to how on Java optionals you have `orElse(val)` and `orElseGet(() -> val)`.  But I guess that it's not a very common use-case seeing how it hasn't come up before.

iyefrat:
[...]

PR #593, opened 2024-10-10
feat: Add unwrapOrElse

sya-ri:
Similar to unwrap, but returns default value by function: #587

```typescript
const myResult = err('Oh noooo')

const multiply = (value: number): number => value * 2

const unwrapped: number = myResult.map(multiply).unwrapOrElse(() => 10)
```

```typescript
const unwrapped: number = await errAsync(0).unwrapOrElse(() => 10)
// unwrapped = 10
```

Useful for lazily evaluating defaults or performing processing.

unwrapOrElse exists in Rust:
https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else

### DIFF
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -192,13 +192,7 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
   }
 
   unwrapOrElse<A>(err: (e: E) => A): Promise<T | A> {
-    return this._promise.then((res) => {
-      if (res.isErr()) {
-        return err(res.error)
-      }
-
-      return res.value
-    })
+    return this._promise.then((res) => res.unwrapOrElse(err))
   }
 
   /**
--- a/src/result-async.ts
+++ b/src/result-async.ts
@@ -191,6 +191,16 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
     return this._promise.then((res) => res.unwrapOr(t))
   }
 
+  unwrapOrElse<A>(err: (e: E) => A): Promise<T | A> {
+    return this._promise.then((res) => {
+      if (res.isErr()) {
+        return err(res.error)
+      }
+
+      return res.value
+    })
+  }
+
   /**
    * Emulates Rust's `?` operator in `safeTry`'s body. See also `safeTry`.
    */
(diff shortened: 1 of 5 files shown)


src/result-async.ts
```diff
@@ -191,6 +191,16 @@ export class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
     return this._promise.then((res) => res.unwrapOr(t))
   }
 
+  unwrapOrElse<A>(err: (e: E) => A): Promise<T | A> {
+    return this._promise.then((res) => {
+      if (res.isErr()) {
+        return err(res.error)
```
macksal, line 197: reading this was a little confusing because of the `err` function in the library :)

macksal (approved):

src/result-async.ts
sya-ri: Indeed. Rereading, I was confused too 😆 

I made it better 😉 

https://github.com/supermacro/neverthrow/pull/593/commits/0f34663e2f2310e41affcb56d14b1bd9ac823a5a

src/result.ts
```diff
@@ -447,6 +459,10 @@ export class Err<T, E> implements IResult<T, E> {
     return v
   }
 
+  unwrapOrElse<A>(err: (e: E) => A): T | A {
+    return err(this.error)
+  }
```
macksal, line 464: I still find the naming `err` to be a bit confusing, but now I see that it's used everywhere for this type of argument 🤷🏼 e.g. in match directly below. no reason to break that pattern...

tests/index.test.ts
```diff
@@ -204,6 +204,11 @@ describe('Result.Ok', () => {
     expect(okVal.unwrapOr(1)).toEqual(12)
   })
 
+  it('unwrapOrElse and return the Ok value', () => {
+    const okVal = ok(12)
+    expect(okVal.unwrapOrElse(() => 1)).toEqual(12)
+  })
```
macksal, line 210: I think it's part of unwrapOrElse's contract to _not_ call the callback for an Ok value.

It could make sense to test that (using `jest.fn` and `expect(...).not.toHaveBeenCalled()`)

supermacro:
This has already existed for a long time:

https://github.com/supermacro/neverthrow?tab=readme-ov-file#resultunwrapor-method

... or am I missing something here?

PR #610, opened 2024-11-12
feat: Add a function that combines multiple results and returns an object

Karibash:
Hi @supermacro ,

First of all, I would like to thank you for creating such a wonderful library. I have been using neverthrow in a production environment for about a year and I am very satisfied with it.

In this PR, I am adding a function that combines multiple Results, passed as an object, into a single Result. This can be very useful for functional programming as shown below:

```ts
const createTaskId: (id: string) => Result<TaskId, Error>;
const createTaskContent: (content: string) => Result<TaskContent, Error>;

const task = Result.struct({
  id: createTaskId('id'),
  content: createTaskContent('content'),
});
```

To do the same thing with the current API, you need to use `map` as shown below. This requires you to remember which data is in which position in the array, which is not very convenient.

```ts
const createTaskId: (id: string) => Result<TaskId, Error>;
const createTaskContent: (content: string) => Result<TaskContent, Error>;

const values = Result.combineWithAllErrors([
  createTaskId('id'),
  createTaskContent('content'),
]);

const task = values.map(([id, content]) => ({ id, content }));
```


macksal:
Faced this exact pattern before, and using ResultAsync.combine is very awkward. +1 for the feature.

My one question is, does this really need to be a different method, or would it be possible to just overload `combine`/`combineWithAllErrors` to accept either an array or an object and return the appropriate type?

I am thinking of the `combineLatest` function in rxjs which has that exact overload. Pass an array of observables, get an observable of arrays. Pass an object of observables, get an observable of objects.

We could have similarly (for `Result.combine`):
- pass an array of results, get a result of array
- pass an object of results, get a result of object

overloads can be messy, but I'm more favourable of them in library code where the user is more likely to see these two cases as equivalent, and it avoids the need for them to remember any difference in behaviour between the two if they are truly analogous.

Issue #594, opened 2024-10-12
.match with functions that return Result<., .>


timvandam:
Currently it does not seem possible to handle both `Ok` and `Err` at the same time. Note that this behavior is different from `.andThen(...).orElse(...)` or `.orElse(...).andThen(...)` as the first of the two chained functions may have its return value handled by the second of the two. As far as I am aware there is no neat way of handling this currently. `.match` comes close, but returns a `Promise` rather than `ResultAsync` when applied to `ResultAsync`.

I propose adding the following:
```ts
declare class ResultAsync<T, E> implements PromiseLike<Result<T, E>> {
  fork<A, B>(ok: (t: T) => Result<A, B> | ResultAsync<A, B>, _err: (e: E) => Result<A, B> | ResultAsync<A, B>): ResultAsync<A, B>
}
```

timvandam:
The pattern I have adopted in the meantime is:

```ts
fromSafePromise(x.match(asyncFn1, asyncFn2).then(id))
```

Where `id` is an identity function that serves to flatten the promise returned by the match branch handler

paduc:
Hi @timvandam,

I’m having a hard time understanding what you want to do. Could you provide a more complete example ?

timvandam:
> Hi @timvandam,
> 
> I’m having a hard time understanding what you want to do. Could you provide a more complete example ?

What I’m trying to achieve is behavior analogous to Promise.then with 2 arguments: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#onrejected. This is not the same as chaining .then and .catch (and similarly .andThen and .orElse can nit be applied to achieve the same behavior in a simple way)

You provide two functions that handle the ok/err cases and return a (Async)Result

paduc:
Thanks @timvandam. Reading [this](https://stackoverflow.com/questions/24662289/when-is-thensuccess-fail-considered-an-antipattern-for-promises) also helped me grasp the topic.

Wouldn't the best solution be to mimick `Promise.then` and accept a second function argument to `Result.andThen` ?

timvandam:
@paduc Yes that would be fine too. I personally would prefer `fork` because its much clearer in my opinion. And neverthrow does not necessarily follow the Promise naming (e.g. by using `orElse` rather than something like `catch`) so I think it would be nicer

macksal:
I like this because it kind of completes the "toolbox".

For mapping to a result:
[...]

PR #621, opened 2025-03-03
feat: fork()

timvandam:
https://github.com/supermacro/neverthrow/issues/594

fork() allows you to handle Ok<.> and Err<.> at once. This is similar to Promise:then(a, b), which is not the same as Promise .then(a).catch(b) nor .catch(b).then(a) (see image https://stackoverflow.com/a/24663315)

While match() could already be used to achieve this for Result<A, B> by simply returning a Result<T, U> in either branch, this approach falls apart for ResultAsync, as match() makes the return type to be a promise. Fork() is especially useful in this case, as it allows you to chain methods available on ResultAsync.


timvandam:
Hi @supermacro! Do you think you could review this sometime soon? Thanks!

timvandam:
@supermacro any chance you could review this / let me know if this has no chance of making it in?

Issue #628, opened 2025-03-28
Add `toSerializable` method for Result and AsyncResult


abdelkd:
It would be really helpful to add a function `toSerializable` for `Result` and `AsyncResult` that gets ride of functions (map, mapErr, unwrapOr...etc) and changes the `isOk` and `isErr` functions into boolean values. so the result object will be something like this.

```typescript
type Obj<T> = {
  isOk: boolean;
  isErr: boolean;
  value: T;
}
```

This may look useless but its really needed when it comes to using `neverthrow` with cached server actions (maybe server actions too?) since all the result would get serialized you cannot use functions especially `isOk` and `isErr`. 

an example would look like this: 
```typescript
const serverAction = () => {
  return ok({ env: 'action' }).toSerializable()
}
```

I am more that happy making changes and filing a PR.

antspy:
+1 Yes this would be really helpful. 

Perhaps you could even implement `.toJson()` instead, so that when react calls `JSON.stringify` to send results from server to client, it will automatically do the right thing. So one could just return a `Result<>` instance from the server and get a `Result<>` instance on the client :)

cipherlogs:
I suspect the issue may be related to serialization.

My server action returns a `Result`, but when consumed in a client component, the `ok` appears as an empty array while the `Err` is an infinitely nested array.

I guess server actions can't return Promise<Result> or AsyncResult, right?

abdelkd:
as @antspy stated, [turns out](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) exposing a `.toJSON()` method would allow to define what the serialized format would look like. 

in this case, I think adding `.toJSON` method that omits most fields and methods except `value` and add `isOk` and `isErr` booleans. Does someone have any idea of other fields that would be helpful?

oosawy:
While automatic and implicit serialization using toJSON is great, I think it would be nice to also have `toPlain` and `fromPlain` methods that explicitly convert to and from plain objects. This way, you can work with plain objects directly without converting them to JSON strings.

oosawy:
> [@oosawy](https://github.com/oosawy) You mean instead of toJSON use `toPlain` and `fromPlain`?

Exactly. That said, this doesn't prevent the use of toJSON. In protocols like RSC, where objects are serialized directly, using a plain object can be more suitable than a JSON string in some cases.

abdelkd:
[...]

PR #636, opened 2025-04-23
Add toJSON methdo for serialization. 

abdelkd:
Since `ok` and `err` are classes, when they're returned from server action to a client component the result would loose all methods including `isOk` and `isErr`.

This PR is related to the issue #628 

I am thinking of adding the same way to `ResultAsync` but then on client it would become just like `Result``.

This currenly includes `value` in case of Ok and `error` in case of Err, `isOk` and `isErr` booleans.

### DIFF
--- a/src/result.ts
+++ b/src/result.ts
@@ -147,6 +147,15 @@ interface IResult<T, E> {
   isErr(): this is Err<T, E>
 
   /**
+   * This function is used by `JSON.stringify` to serialize the result
+   * without loosing isOk and isErr by turning them into booleans
+   *
+   * @returns `{ error, isOk, isErr }` if the result is an `Err` variant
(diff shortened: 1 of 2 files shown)


abdelkd:
@supermacro @m-shaka I have an idea of `safeParse` which would parse object in a type safe way, maybe something like this:

```typescript
import { ok, safeParse } from "neverthrow"

const okVal = ok(123)
const unsafeParsed = JSON.parse(JSON.stringify(okVal)) // result is of type any

const typedParsed = safeParse(JSON.stringify(okVal))
if (typedParsed.isErr) {
  typedParsed.error // available here
  throw new Error('Something went wrong')
}

typedParsed.isOk // true
typedParsed.value // 123
```

maxmellen:
Any reason this is being held back?

PR #644, opened 2025-05-06
Feat/add and or methods

kevinautin:
Adds the `and` and `or` methods for better ergonomics, reducing boilerplate.

In short: 
1. `.and(v)`  can be used instead of `.map(() => v)` 
2. `.or(v)` can be used instead of `.orElse(() => ok(v))`


kevinautin:
My code editor automatically formatted some large chunks of the README. 
As far as I can tell, it didn't change anything in terms of how it rendered, but please let me know if you would like me to revert those to their previous state.

PR #658, opened 2025-08-10
Add `this` context support to `safeTry` function

### LINKED ISSUE #632: Feature Request: Allow Passing `this` Context to `safeTry`
I would like to propose a feature enhancement for `neverthrow`'s `safeTry` that enables passing the `this` context to the function being executed. This change would mimic the interface of effect-ts and allow patterns like the following:

```ts
compute = Effect.gen(this, function* () {
  const n = this.local + 1;

  yield* Effect.log(`Computed value: ${n}`);

  return n;
});
```
See: https://effect.website/docs/getting-started/using-generators/#passing-this

**Current Behavior**  
At the moment, `safeTry` does not allow the caller to bind the `this` context, making it difficult to use in object-oriented patterns where instance properties are required inside the callback.

**Expected Behavior**  
The expected behavior is for `safeTry` to accept an optional `this` context and invoke the callback with that context. For example, a possible usage could be:

```ts
// In a class definition
myFunc(): Result<number, string> {
    return safeTry<number, string>(this, function*() {
        return ok(
            (yield* this.mayFail1()
                .mapErr(e => `aborted by an error from 1st function, ${e}`))
        )
    })
}
```

**Benefits**  
- **Enhanced usability:** Allowing `this` binding makes `safeTry` more versatile, especially when working within class-based structures.
- **Improved integration:** This feature would simplify the integration of `neverthrow` into codebases that rely on object-oriented design and effectful computations.

I would appreciate any feedback on this proposal. Thank you for considering this feature request!

yasaichi:
This PR implements support for passing `this` context to the generator function in `safeTry`, following the established pattern already implemented in similar libraries like [Effect](https://effect.website/docs/getting-started/using-generators/#passing-this) and [typescript-result](https://www.typescript-result.dev/chaining-vs-generator-syntax#this-context).

Closes #632

### Usage

```typescript
class MyClass {
  private multiplier = 2;
  
  getValue(): number {
    return 10;
  }
  
  myMethod(): Result<number, string> {
    return safeTry(this, function* () {
      const value = this.getValue();
      return ok(value * this.multiplier);
    });
  }
}
```

### Why This Matters

This feature enhances the ergonomics of `safeTry` when used in class-based designs, eliminating the need for manual binding to preserve context (e.g. `that`). This brings neverthrow in line with other modern Result/Effect libraries that have already adopted this pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1582,7 +1582,7 @@ declare function mayFail1(): Result<number, string>;
 declare function mayFail2(): Result<number, string>;
 
 function myFunc(): Result<number, string> {
-    return safeTry<number, string>(function*() {
+    return safeTry(function*() {
         return ok(
             // If the result of mayFail1().mapErr() is an `Err`, the evaluation is
             // aborted here and the enclosing `safeTry` block is evaluated to that `Err`.
@@ -1610,8 +1610,8 @@ You can also use [async generator function](https://developer.mozilla.org/en-US/
 declare function mayFail1(): Promise<Result<number, string>>;
 declare function mayFail2(): ResultAsync<number, string>;
 
-function myFunc(): Promise<Result<number, string>> {
-    return safeTry<number, string>(async function*() {
+function myFunc(): ResultAsync<number, string> {
+    return safeTry(async function*() {
         return ok(
             // You have to await if the expression is Promise<Result>
             (yield* (await mayFail1())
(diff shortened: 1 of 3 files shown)


yasaichi:
@supermacro What can I do for you to merge this PR? Thanks.

